PR #86, opened 2016-05-30
Font awesome updates

lortza:
In addition to some html and css changes to resolve the fontawesome issue, this also does a little whitespace cleanup. @mikevallano I'd like a second set of eyes on this just incase anything looks a little fishy. You can merge and push to heroku if you think it's good to go.

### DIFF
--- a/app/views/lists/index.html.erb
+++ b/app/views/lists/index.html.erb
@@ -4,7 +4,7 @@
   <h1>My Movie Lists</h1>
 
   <div class="pull-right">
-     <p class="button-main"><%= link_to '<i class="fa fa-plus"></i> New List'.html_safe, new_user_list_path(current_user) %></p>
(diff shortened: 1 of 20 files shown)


app/views/lists/index.html.erb
```diff
@@ -4,12 +4,9 @@
   <h1>My Movie Lists</h1>
 
   <div class="pull-right">
-     <p class="button-main"><%= link_to ' New List', new_user_list_path(current_user), id: "new_list_link_list_index", class: "fa fa-plus" %></p>
```
mikevallano: @lortza plz keep the ids on these

app/views/lists/index.html.erb
```diff
@@ -4,12 +4,9 @@
...
+     <p class="button-main"><%= link_to '<i class="fa fa-plus"></i> New List'.html_safe, new_user_list_path(current_user) %></p>
```
mikevallano: @lortza other id to keep

PR #128, opened 2017-12-28
Condenses navbar options to fit the space better

lortza:
Old navbar has an annoying line break
screenshot 2017-12-28 00 00 33

New navbar fits on one line, not perfectly, but better
screenshot 2017-12-28 00 00 21

Moved nav items as such:
- "My Lists" -> "Lists"
- "My Movies" -> Lists/Recently Watched
- "Public Lists" -> Lists/Public Lists
- "Lists/Manage My Lists" -> signed in as:... / Manage My Lists since we rarely use this feature
- "Lists/Add a List" -> removed from nav, now just a link on the My Lists index page
 
screenshot 2017-12-28 00 02 18

screenshot 2017-12-28 00 02 25

Be advised... I didn't run the test suite or anything. :0

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -347,9 +347,3 @@ DEPENDENCIES
   web-console (~> 2.0)
   webmock
   will_paginate
-
-RUBY VERSION
-   ruby 2.2.2p95
-
-BUNDLED WITH
-   1.15.1
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -347,3 +347,9 @@ DEPENDENCIES
   web-console (~> 2.0)
   webmock
   will_paginate
+
+RUBY VERSION
+   ruby 2.2.2p95
+
+BUNDLED WITH
+   1.15.1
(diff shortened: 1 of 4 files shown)


mikevallano:
👍 Nice!

Two items:
- [ ] You may not be able to successfully run the tests as they are in a janked state, but can you try to run `lists_feature_spec.rb`? It looks like there are a couple of tests that will be out of date with these changes. If you get an error unrelated to the actual tests, just don't worry about it. If you get a failing test or two related to these changes, can you please fix them?

- [ ] The Gemfile.lock changed. I remember running into this before. I think it's because we're using different versions of bundler. We should probably get in sync. But for now, can you just omit that change from the PR since it's unrelated to the style changes?

lortza:
I've resolved the bundler version error for this PR. We both need to get on bundler version `1.16.1` now and then stay up to date to avoid this in the future. Running `gem install bundler` should get us both up to date.

I'm working on the tests right now.

Issue #138, opened 2018-09-01
Show age of actors in actor profile view


mikevallano:
We show the actor's date of birth, but not age. 

It might also be cool to show the person's age for each movie, but that might also be overkill in the view.

lortza:
This feature is WIP.

- [ ] Write tests for MoviePersonProfile class

- [x] Refactor age and minute count display into Time helper

- [x] Write tests for new Time helper & update tests for movie_helper

- [x] Remove unused instance variables in the tmdb_handler 

- [x] Rename @person_bio to @person_profile across views, controllers, and models

PR #142, opened 2018-09-03
138 actors age

lortza:
This PR addresses issue https://github.com/mikevallano/tmdb-moviequeue/issues/138 which includes the age in a movie person's bio (a movie person is an actor or a director)

screenshot 2018-09-03 12 52 43

This branch includes tests for all new features and does a little RuboCop cleanup with the surrounding code. Currently all tests except feature specs are passing.


app/helpers/movies_helper.rb
```diff
@@ -67,9 +67,7 @@ def full_cast_link(movie)
   end
 
   def runtime_display(movie)
-    if movie.runtime.present?
-      movie.runtime % 60 == 0 ? "#{movie.runtime/60}hr" : "#{movie.runtime/60}hr #{movie.runtime % 60}min"
-    end
+    DateAndTimeHelper.display_time(movie.runtime) if movie.runtime.present?
   end
```
lortza: The new `DateAndTimeHelper` handles all calculations about dates and times. That is why all of this logic was moved from the `movies_helper`. This new file refactor happened when we needed to calculate a person's age.

app/models/movie_person_profile.rb
```diff
@@ -0,0 +1,59 @@
...
+  validates :person_id, :name, :bio, :birthday_and_age, :profile_path, presence: true
+
+  WIKIPEDIA_CREDIT = {
+    starting: 'From Wikipedia, the free encyclopedia.',
+    trailing: 'Description above from Wikipedia.',
+    standard: 'Bio from Wikipedia.'
+  }.freeze
+
+  class << self
+    def parse_result(result)
```
lortza: In case you haven't seen this yet (it's new to me!), using a section like:
```
class << self
  def something
  end
end
```
is a tidier way to organize a bunch of class-level methods like:
```
def self.something
end

def self.something_else
end
```
This way you don't have to have a bunch of loose `def self.something`s which could get accidentally mixed in with your other methods.

app/helpers/date_and_time_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def years_since_date(date)
+      today = Time.zone.today
+      today.year - date.year - allowance_for_same_date(date, today)
+    end
+
+    private
+
+    def allowance_for_same_date(date, today)
```
mikevallano: try to come up with a better name

app/helpers/date_and_time_helper.rb
```diff
@@ -0,0 +1,34 @@
+module DateAndTimeHelper
+  class << self
+    MINUTES_IN_AN_HOUR = 60
+
+    def display_time(minutes)
+      return nil if minutes.blank?
+
+      full_hour?(minutes) ? display_time_in_hours(minutes) : display_time_in_hours_and_minutes(minutes)
+    end
```
lortza: I've replaced the guard clause syntax with a ternary since there was only one action following that clause.

app/helpers/date_and_time_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+    def years_since_date(date)
+      today = Time.zone.today
+      today.year - date.year - (date_occurred_yet_this_year?(date, today) ? 0 : 1)
+    end
+
+    private
+
+    def date_occurred_yet_this_year?(date, today)
+      today.month > date.month || (today.month == date.month && today.day >= date.day)
+    end
```
lortza: I changed the abstraction here so the `years_since_date` and `date_occurred_yet_this_year?` methods are easier to follow.

mikevallano: 👍 Sweet.

Issue #157, opened 2019-08-22
Upgrade to rails 5


mikevallano:
It's about time.

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0

PR #158, opened 2019-08-22
156: Tick version of coveralls to remove rest-client dependency

mikevallano:
The next minor version of `coveralls` does not have `rest-client` as a dependency, so this upgrades to that version. 

We should probably upgrade to Rails 5, so I created an issue for it: https://github.com/mikevallano/tmdb-moviequeue/issues/157


lortza:
The build failure is interesting. I don't know why it can't find postgres now when it could in the past. But I'm going to go investigate a hunch I have... brb

Just curious, did you run the tests locally?

lortza:
> But I'm going to go investigate a hunch I have... brb
Nope. Didn't work. :(

mikevallano:
@lortza I rebased master into this branch (since master now has the fix for the build), and force pushed. I think you previously pushed to this branch because my `--force-with-lease` was rejected. Just giving you a heads-up. 

If the build passes, plz review.

PR #169, opened 2019-11-30
154: Sort lists by name when adding to a list

### LINKED ISSUE #154: order Lists by name
Lists appear an alpha order (ascending) in the nav menu, but not in the modal or show page. We'd like to have a consistent ASC ordering in all places.

mikevallano:
Closes #154 

This whole sitch needs to be refactored, but for now, it will order the lists by name.

### Screenshots

**When movie is not on a list yet**
Screen Shot 2019-11-30 at 9 46 33 AM

**When movie is on list(s)**
Screen Shot 2019-11-30 at 9 46 46 AM

**Show page**
Screen Shot 2019-11-30 at 9 57 36 AM

**Lists in nav (unchanged)**
Screen Shot 2019-11-30 at 9 49 34 AM

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -62,11 +62,11 @@ class User < ActiveRecord::Base
   end
 
   def lists_except_movie(movie = nil)
-    if movie.present? && Movie.exists?(tmdb_id: movie.tmdb_id)
-      @movie_lists = Movie.find_by(tmdb_id: movie.tmdb_id).lists.by_user(self)
-      @lists_except_movie = (all_lists - @movie_lists)
+    if movie.present? && existing_movie = Movie.find_by(tmdb_id: movie.tmdb_id)
+      (all_lists - existing_movie.lists.by_user(self))
+        .sort_by { |list| list.name.downcase }
     else
-      all_lists
+      all_lists_by_name
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -62,8 +62,8 @@ class User < ActiveRecord::Base
   end
 
   def lists_except_movie(movie = nil)
-    if movie.present? && existing_movie = Movie.find_by(tmdb_id: movie.tmdb_id)
-      (all_lists - existing_movie.lists.by_user(self))
+    if movie.present? && movie.in_db
+      (all_lists - movie.lists.by_user(self))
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -62,11 +62,11 @@ def all_listings
   end
 
   def lists_except_movie(movie = nil)
-    if movie.present? && Movie.exists?(tmdb_id: movie.tmdb_id)
-      @movie_lists = Movie.find_by(tmdb_id: movie.tmdb_id).lists.by_user(self)
-      @lists_except_movie = (all_lists - @movie_lists)
+    if movie.present? && existing_movie = Movie.find_by(tmdb_id: movie.tmdb_id)
+      (all_lists - existing_movie.lists.by_user(self))
+        .sort_by { |list| list.name.downcase }
```
mikevallano: The `sort_by` bit is the only actual change. the other stuff was just refactoring.

The `sort_by` is needed because of the existing jank setup that is doing two different queries (checking a lists's owner, as well as any lists the user is a _member_ of). It returns an array, not an ActiveRecord relationship, hence the need to sort it this way.

This can be improved by upgrading to Rails 5 and using the ActiveRecord `or`. But probably a better option is to _always_ use the `memberships` join table (links users to lists), which would simplify a lot. I think. I'd have to look more into it.

lortza: Is `existing_movie = Movie.find_by` supposed to be `=` or `==`? If `=`, is there any chance we can move the assignment of `existing_movie` into another method? Something like 
```
def existing_movie(movie)
  existing_movie = Movie.find_by(tmdb_id: movie.tmdb_id)
end
```

mikevallano:
@lortza Yeah, the whole thing is pretty ugly, but I improved it a bit in the [last commit](https://github.com/mikevallano/tmdb-moviequeue/pull/169/commits/9a1ba6364753350d4d0a808105e9e177b39f9140).

The thing to account for is when you search for a movie, it returns results from the API. Some of those movies are persisted in the database (`in_db`), and some are not. 

So now, it'll just do the `in_db` check and eliminate the other query or the weird assigning I was doing.

Let me know if you think this is passable for now, or if you'd rather handle this as part of a bigger refactor.

Issue #173, opened 2019-12-23
Clean up duplication in forms for adding a new listing


mikevallano:
`app/views/movies/_movie_add_to_first_list.html.erb` and `app/views/movies/_movie_add_to_another_list.html.erb` are nearly identical, and there's logic around it in `app/views/movies/_movie_modal_innard.html.erb`.

This could probably be cleaned up as it seems the only difference is the placeholder text, but I don't know if there are other implications.

PR #174, opened 2019-12-23
153: Add autocomplete when adding a movie to a list

### LINKED ISSUE #153: add autocomplete to List dropdown on modal and show page
It is a PIA to use eyeballs to look at all of the list names when selecting one for a movie. Autocomplete would make this much better. It may involve initializing the autocomplete library in the js.

mikevallano:
Closes #153.

Adds autocomplete when adding a movie to a list.

While working on this, I noticed there's duplication in forms, when adding to a list for the first time vs future times. That means there's duplication in partials and logic. I'm not sure the implications, so I created a separate issue for refactoring it: https://github.com/mikevallano/tmdb-moviequeue/issues/173

As for other shitty code, there's also duplication because autocomplete needs to be triggered on initial page load, when the partial is rendered after creating a new listing, and when the modal is opened.

There have to be better ways of handling it, but 🤷‍♂ . Ideally it'd be good to refactor to use react or vue, but for now, this works.

### Screenshots
add-to-list-autocomplete
add-to-list-autocomplete2


app/assets/javascripts/z.js
```diff
@@ -32,19 +42,27 @@ $(document)
     }
   });
 
+const handleAddToListAutocompleteSelect = function(event, ui) {
+  $("#list_autocomplete_val").val(ui.item.id);
+  $('form#new_listing').submit();
+}
 
-    // This hides the 'year select' field until a year is selected
```
mikevallano: this was just old, unused commented out code

app/views/movies/_movie_add_to_another_list.html.erb
```diff
@@ -1,10 +1,17 @@
 
   <% if current_user.lists_except_movie(movie).count > 0 %>
     <%= form_for(Listing.new, class: "form-class", id: "modal-add-to-other-list", remote: true) do |f| %>
-      <%= f.collection_select :list_id, current_user.lists_except_movie(movie), :id, :name, { prompt: "Add to Another List" }, {class: "form-control"} %>
+      <%= text_field_tag  'list_dropdown',
```
mikevallano: this and the other form are the instances of duplication mentioned in the PR description

app/views/movies/_movie_modal_innard.html.erb
```diff
@@ -7,7 +7,8 @@
 
         <% if movie.in_db && movie.lists.by_user(current_user).present? %>
 
-        <div class="partial-spacer">
+        <div class="partial-spacer ui-front">
+          <!-- ui-front is needed for jquery autocomplete in the modal -->
```
mikevallano: The reason this is needed is explained in this SO post: https://stackoverflow.com/a/25021925. Basically, in the boostrap modal, this class needs to be defined on the parent of the element that will use jquery autocomplete.

app/assets/javascripts/z.js
```diff
@@ -32,19 +42,27 @@ $(document)
...
 
-// the final closer
+function showAllListsForAutocomplete() {
+  // hack to show all lists when initially clicking
+  // in the field. this triggers a backspace keypress
+  let evt = jQuery.Event('keydown');
+  evt.which = 8; // backspace
+  evt.keyCode = 8
+  $('#add-to-list-dropdown').trigger(evt)
+}
```
lortza: that does seem janky, 🤷‍♀but it's working 😃

app/views/movies/_movie_modal_innard.html.erb
lortza: Dag. Nice work in figuring that out!

PR #179, opened 2020-01-06
178 remove modal for mobile

### LINKED ISSUE #178: Remove index modal on mobile
When I'm on my phone, I'm always clicking through the modal to get to the show page for a movie. It makes sense to me to remove the modal for mobile. I think the modal is still useful for desktop view, so i'd rather not remove it completely right now. Tucking it behind a media query seems like the easiest approach at the moment.

lortza:
## Screenshots
Mobile view links directly to show page
Kapture 2020-01-05 at 18 52 46

Tablet and larger still use the modal
Kapture 2020-01-05 at 18 59 22


mikevallano:
Nice work! I'm sure it was tough navigating through that mess of code, but I think this will make your mobile experience much better.

> I also added a PR template to make our lives easier when writing up a PR

Grassy!

> We've got a lot of complicated views and routes going on

Indeed. There's so much room for refactoring. I didn't delve too deep, but I think the reason for the modal route is because it's basically a different view and needs to do queries.  

I'm gonna leave a comment, but other than that this looks good!

app/views/movies/_movie_partial_loop.html.erb
```diff
@@ -4,17 +4,21 @@
...
+        </p>
 
-        <div id="movies_partial_<%= "#{movie.tmdb_id}" %>">
+        <div id="modal-trigger">
+          <% if @list.present? %>
+            <p><%= link_to image_for(movie), movie_modal_path(tmdb_id: movie.tmdb_id, list_id: @list.id), remote: true, data: { target: "#myModal_#{movie.tmdb_id}" }, id: "modal_link_#{movie.tmdb_id}" %></p>
+          <% else %>
+            <p><%= link_to image_for(movie), movie_modal_path(tmdb_id: movie.tmdb_id),                    remote: true, data: { target: "#myModal_#{movie.tmdb_id}" }, id: "modal_link_#{movie.tmdb_id}" %></p>
+          <% end %>
+        </div>
```
mikevallano: It looks like this whole if/else statement is only determining if it should pass `list_id` as a param.

So I think we can simplify it by just using safe navigator on `@list&.id`, and remove the if/else (I tested it locally):

```html
<div id="modal-trigger">
   <p><%= link_to image_for(movie), movie_modal_path(tmdb_id: movie.tmdb_id, list_id: @list&.id), remote: true, data: { target: "#myModal_#{movie.tmdb_id}" }, id: "modal_link_#{movie.tmdb_id}" %></p>
</div>
```
I know you didn't edit that bit of code, but since the file is being edited, I was thinking we may as well clean it up. I think it could be done in this PR, but if you'd rather leave it for a separate issue, that's fine w/ me.

app/assets/stylesheets/movies.scss
```diff
@@ -395,6 +397,12 @@ article.tile-cover-pic {
     } /* .col-xs-9 */
   } /* .modal */
 
+  // article.tile-cover-pic {
+  //   #modal-trigger {
+  //     display: none;
+  //   }
+  // }
+
```
mikevallano: Did you mean to leave this here? If so, that's fine; I'm just checking.

lortza: Nooooope! I'll cut that back out.

app/views/movies/_movie_partial_loop.html.erb
lortza: I'm cool with adding this change since you tested it locally. I wasn't sure if the presence of the param key with a nil value would cause issues. But if we're cool, i'm game.

lortza:
I'll make those changes, do some browser testing, and push them.

lortza:
Okey doke, i've pushed those changes. Curiously though, while browser testing, i noticed that the search function does not work locally on this branch or on `master`.

lortza:
Do you want to do merge & deploy or shall i?

PR #188, opened 2020-02-29
WIP this fixes the broken link on mobile

### LINKED ISSUE #180: Can't get to show page from mobile
After I removed the modal from the portrait mobile view (via media query), when i clicked on a movie cover, i got a 404 when trying to go to the show page. I rotated my phone to landscape where the modal could still work, clicked on the link to `show` and it worked. My guess is that we're missing some sort of call to the API to build that info before we get to the `show` page.

lortza:
## Problems Solved
After I removed the modal from the portrait mobile view (via media query), when i clicked on a movie cover, i got a 404 when trying to go to the show page. I rotated my phone to landscape where the modal could still work, clicked on the link to show and it worked. 

My guess was that we were missing some sort of call to the API to build that info before we get to the show page. To work around that, I used a route that let me pass in the tmdb id:
```ruby
link_to image_for(movie), movie_path(movie)
```

With this workaround, I'm not sure we're using the `movies/show` page anymore. 🤷‍♀ This may not be the most sound approach.

## Screenshots
Before
before

After
after

### DIFF
--- a/app/views/movies/_movie_partial_loop.html.erb
+++ b/app/views/movies/_movie_partial_loop.html.erb
@@ -12,7 +12,8 @@
         </p>
 
         <p id="modal-trigger-mobile">
-          <%= link_to image_for(movie), movie_path(movie) %>
+          <%#= link_to image_for(movie), movie_path(movie) %>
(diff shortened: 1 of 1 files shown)


lortza:
This is fixed in #232

Issue #195, opened 2020-05-17
Once images are fixed, determine ongoing schedule for rake task and schedule it Heroku


mikevallano:
We want to keep data in sync w/ TMDB, so we should decide how frequently, and in what batch size we want to do it. 

Maybe once a month?

lortza:
Notes from our conversation yesterday:

I like the idea of once a month. We can look at our data to see what our average monthly `movies` table additions are, or we can just take a guess (60?). 

For the initial back-data, we'll query for the oldest `updated_at` with a limit of 300(?) and do that over the course of X days until that number of remaining un-updated-recently records drops below X.

mikevallano:
Closed with #202

PR #196, opened 2020-05-17
193: Get updated tmdb data

### LINKED ISSUE #193: TMDB images are broken
TMDB images are broken and this kind of wrecks the whole way our app works. 
Screenshot 2020-05-10 09 11 05

As far as we know, the problem is on TMDB's side. We should handle this in some way so that we can still use the app.

mikevallano:
## Related Issues & PRs
Closes #193

Adds a rake task to get latest data from TMDB and update the data in our DB accordingly. 

FYI, as of 5/17/20, there are 1,752 movies in our DB.

## Problems Solved
* The primary goal is to fix the broken images. But it's also good to regularly update the movie data from TMDB. There's a separate issue for that here: https://github.com/mikevallano/tmdb-moviequeue/issues/195

## Screenshots
N/A

## Things Learned
* The code written 4 years ago isn't great 😄 . 
* Scheduling tasks in Heroku is a thing (grassy @lortza )

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'sdoc', '~> 0.4.0', group: :doc
 gem 'devise'
 gem 'figaro'
 gem 'friendly_id'
+gem 'httparty'
 gem 'rails_12factor', group: :production
 gem 'will_paginate'
 gem 'coveralls', '0.8.11', require: false
--- a/lib/tasks/update_tmdb_data.rake
+++ b/lib/tasks/update_tmdb_data.rake
@@ -1,8 +1,8 @@
 # To run the rake task: rake "tmdb_data:refresh"
 # To run from console: Rake::Task["tmdb_data:refresh"].invoke
 
-TIME_FRAME = 1.week.ago
-BATCH_SIZE = 50
+TIME_FRAME = (ENV['tmdb_refresh_time_frame_days'] || 7).to_i.days.ago
+BATCH_SIZE = (ENV['tmdb_refresh_batch_size'] || 50).to_i
 
 namespace :tmdb_data do
   desc 'refresh tmdb_data'
--- a/lib/tasks/update_tmdb_data.rake
+++ b/lib/tasks/update_tmdb_data.rake
@@ -7,19 +7,18 @@ BATCH_SIZE = (ENV['tmdb_refresh_batch_size'] || 50).to_i
 namespace :tmdb_data do
   desc 'refresh tmdb_data'
   task :refresh => :environment do
-    tmdb_ids = Movie
+    movies = Movie
       .where('updated_at < ?', TIME_FRAME)
       .order(updated_at: :asc)
       .limit(BATCH_SIZE)
-      .pluck(:tmdb_id)
 
-    puts "*** Refreshing #{tmdb_ids.count} movies..."
+    puts "*** Refreshing #{movies.count} movies..."
     updated_movies = []
 
-    tmdb_ids.each do |tmdb_id|
-      TmdbHandler.tmdb_handler_update_movie(tmdb_id)
-      puts "Movie refreshed. tmdb_id: #{tmdb_id}"
-      updated_movies << tmdb_id
+    movies.each do |movie|
+      TmdbHandler.tmdb_handler_update_movie(movie)
+      puts "Movie refreshed. tmdb_id: #{movie.tmdb_id}"
+      updated_movies << movie.tmdb_id
       sleep 0.01
     end
     output = "*** Updated_movies: count: #{updated_movies.count}. tmdb_ids: #{updated_movies}"
--- a/lib/tasks/update_tmdb_data.rake
+++ b/lib/tasks/update_tmdb_data.rake
@@ -22,7 +22,11 @@ namespace :tmdb_data do
       updated_movies << tmdb_id
       sleep 0.01
     end
-
-    puts "*** Updated_movies: count: #{updated_movies.count}. tmdb_ids: #{updated_movies}"
+    output = "*** Updated_movies: count: #{updated_movies.count}. tmdb_ids: #{updated_movies}"
+    puts output
+  rescue => error
+    output = "*** Updated_movies: count: #{updated_movies.count}. tmdb_ids: #{updated_movies}"
+    puts "Not all movies uptated. #{output}"
+    raise error
   end
 end
--- a/lib/tasks/update_tmdb_data.rake
+++ b/lib/tasks/update_tmdb_data.rake
@@ -1,5 +1,5 @@
-# rake "tmdb_data:refresh"
-# from console: Rake::Task["tmdb_data:refresh"].invoke
+# To run the rake task: rake "tmdb_data:refresh"
+# To run from console: Rake::Task["tmdb_data:refresh"].invoke
 
 TIME_FRAME = 1.week.ago
 BATCH_SIZE = 50
@@ -17,14 +17,10 @@ namespace :tmdb_data do
     updated_movies = []
 
     tmdb_ids.each do |tmdb_id|
-      res = TmdbHandler.tmdb_handler_update_movie(tmdb_id)
-      if res
-        puts "Movie refreshed. tmdb_id: #{tmdb_id}"
-        updated_movies << tmdb_id
-        sleep 0.01
-      else
-        break
-      end
+      TmdbHandler.tmdb_handler_update_movie(tmdb_id)
+      puts "Movie refreshed. tmdb_id: #{tmdb_id}"
+      updated_movies << tmdb_id
+      sleep 0.01
     end
 
     puts "*** Updated_movies: count: #{updated_movies.count}. tmdb_ids: #{updated_movies}"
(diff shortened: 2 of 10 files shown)


Gemfile
```diff
@@ -15,6 +15,7 @@ gem 'sdoc', '~> 0.4.0', group: :doc
 gem 'devise'
 gem 'figaro'
 gem 'friendly_id'
+gem 'httparty'
```
mikevallano, line 18: Cleaner way of making API requests.

lib/tasks/update_tmdb_data.rake
```diff
@@ -0,0 +1,32 @@
+# rake "tmdb_data:refresh"
+# from console: Rake::Task["tmdb_data:refresh"].invoke
+
+TIME_FRAME = 1.week.ago
+BATCH_SIZE = 50
```
mikevallano, line 5: I wanted to start out with a small sample as a way of testing. If this goes well, I think we could do batches of ~250. 

There are currently 1,752 movies in the DB, so if we ran this every day with 250, we'd update all movies in a week.

I think that's still really conservative since they don't have rate limits, but it's best to be safe.

lib/tmdb_handler.rb
```diff
@@ -81,52 +82,72 @@ def tmdb_handler_add_movie(tmdb_id)
...
-    @movie_to_update = Movie.find_by(tmdb_id: tmdb_id)
-    @movie = MovieMore.tmdb_info(@result)
-
-    @movie_to_update.update_attributes(title: @movie.title, imdb_id: @movie.imdb_id, genres: @movie.genres,
-      actors: @movie.actors, adult: @result[:adult], backdrop_path: @movie.backdrop_path,
-      poster_path: @movie.poster_path, release_date: @movie.release_date, overview: @movie.overview,
-      trailer: @movie.trailer, director: @movie.director, director_id: @movie.director_id,
-      vote_average: @movie.vote_average, popularity: @movie.popularity, runtime: @movie.runtime,
-      mpaa_rating: @movie.mpaa_rating)
+  def self.tmdb_handler_update_movie(tmdb_id)
```
mikevallano, line 85: Here I aimed to be very conservative and report to Sentry if anything unexpected happens.

Also, if anything goes wrong, we'll return nil out of this method, and then in the Rake task, it'll stop iterating and making API calls, just to be safe.

lib/tmdb_handler.rb
```diff
@@ -81,52 +82,72 @@ def tmdb_handler_add_movie(tmdb_id)
...
+    movie = Movie.find_by(tmdb_id: tmdb_id)
+    Raven.capture_message("movie not found in our db: #{tmdb_id}") && return unless movie
+
+    movie_url = "#{BASE_URL}/movie/#{tmdb_id}?api_key=#{ENV['tmdb_api_key']}&append_to_response=trailers,credits,similar,releases"
+    api_result = HTTParty.get(movie_url).deep_symbolize_keys rescue nil
+    Raven.capture_message("API request failed for tmdb_id: #{tmdb_id}") && return unless api_result && api_result[:id] == tmdb_id
```
mikevallano, line 91: If the request to `HTTParty.get(movie_url)` doesn't find a movie, the result looks like:

`{:status_code=>34, :status_message=>"The resource you requested could not be found."}`

I added the rescue just to be safe in case anything unexpected breaks. 

Otherwise, if the hash with the status_message is returned, the `id` won't match, and we'll get notified via Sentry, and the rake task will stop iterating.

lib/tmdb_handler.rb
```diff
@@ -81,52 +82,72 @@ def tmdb_handler_add_movie(tmdb_id)
...
+
+    updated_data = MovieMore.tmdb_info(api_result)
+
+    if movie.title != updated_data.title
+      msg = "Movie title doesn't match. tmdb_id: #{tmdb_id}. Current title: #{movie.title}. Updated title: #{updated_data.title}"
+      Raven.capture_message(msg) && return
+    end
```
mikevallano, line 98: I added this just to be super safe that we're not accidentally borking data.

lib/tmdb_handler.rb
```diff
@@ -81,52 +82,72 @@ def tmdb_handler_add_movie(tmdb_id)
...
+
+    movie.update!(
```
mikevallano, line 100: This is just a syntax cleanup.

lib/tmdb_handler.rb
```diff
@@ -81,52 +82,72 @@ def tmdb_handler_add_movie(tmdb_id)
...
+    )
   end
 
   def tmdb_handler_actor_more(actor_id)
     tmdb_handler_person_detail_search(actor_id)
   end
 
   def tmdb_handler_person_detail_search(person_id)
-    api_bio_url = "https://api.themoviedb.org/3/person/#{person_id}?api_key=#{ENV['tmdb_api_key']}"
+    api_bio_url = "#{BASE_URL}/person/#{person_id}?api_key=#{ENV['tmdb_api_key']}"
```
mikevallano, line 124: This is just a code cleanup so we're not repeating the same base url everywhere. So when it changes from `/3/` to `/4` or whatever, it only needs to be updated in one place.

lib/tasks/update_tmdb_data.rake
lortza: I agree.

mikevallano:
Ok, I made some tweaks: mainly handling potential errors. 

I realized that the rake task will stop if an error is raised, so instead of hacking and returning nil unnecessarily, we can just raise an error if anything unexpected happens, and the rake task will stop, and not continue making API requests, or iterating if the code is busted.

I also realized there was a failing test from your last PR (I fixed it) and that Travis is being fucky: showing `passing` when the build failed, and showing `failed` when the build is passing. I guess we'll need to actually look at the Travis result before merging PRs to be safe.

spec/helpers/movies_helper_spec.rb
```diff
@@ -14,7 +14,11 @@
 
     it 'returns an image tag if the movie has a poster path' do
       allow(movie).to receive(:poster_path).and_return('tester')
-      expect(helper.image_for(movie)).to eq(image_tag("http://image.tmdb.org/t/p/w185#{movie.poster_path}"))
+      expect(helper.image_for(movie)).to eq(
+        image_tag("http://image.tmdb.org/t/p/w185#{movie.poster_path}",
+          title: movie.title,
+          alt: movie.title)
+        )
```
lortza, line 21: Ah, that makes sense. Super weird that Travis didn't give us a clear heads-up on that.

lib/tmdb_handler.rb
lortza: Good call. Borking would be bad.

lib/tmdb_handler.rb
lortza: Yes. Very nice!

lib/tmdb_handler.rb
```diff
@@ -81,52 +88,74 @@ def tmdb_handler_add_movie(tmdb_id)
...
+    movie_url = "#{BASE_URL}/movie/#{tmdb_id}?api_key=#{ENV['tmdb_api_key']}&append_to_response=trailers,credits,similar,releases"
+    api_result = HTTParty.get(movie_url).deep_symbolize_keys rescue nil
+    raise TmdbHandlerError.new("API request failed for tmdb_id: #{tmdb_id}") unless api_result && api_result[:id].to_s == tmdb_id.to_s
+
+    updated_data = MovieMore.tmdb_info(api_result)
+
+    if movie.title != updated_data.title
+      msg = "Movie title doesn't match. tmdb_id: #{tmdb_id}. Current title: #{movie.title}. Updated title: #{updated_data.title}"
+      raise TmdbHandlerError.new(msg)
+    end
```
lortza, line 104: Instead of raising an error, let's just return out of this method. This will skip the update for this particular movie, but allow the rake task to continue to run for other movies:

```ruby
if movie.title != updated_data.title
  puts "Movie title doesn't match. tmdb_id: #{tmdb_id}. Current title: #{movie.title}. Updated title: #{updated_data.title}"

  return
end
```

lib/tasks/update_tmdb_data.rake
```diff
@@ -0,0 +1,28 @@
...
+      .limit(BATCH_SIZE)
+      .pluck(:tmdb_id)
+
+    puts "*** Refreshing #{tmdb_ids.count} movies..."
+    updated_movies = []
+
+    tmdb_ids.each do |tmdb_id|
+      TmdbHandler.tmdb_handler_update_movie(tmdb_id)
+      puts "Movie refreshed. tmdb_id: #{tmdb_id}"
+      updated_movies << tmdb_id
```
lortza, line 22: From what I can tell, the `updated_movies` count will always be the same as the input count. There aren't any conditions in this loop that would prevent a movie from being counted. When we raise an error, it just stops the loop, but doesn't adjust the output count. 

I'm not prioritizing making this count smarter. It seems like it could be a lot of effort for minimal reward.

lortza:
This looks good! I think it is in great shape to do a trial run.

lib/tmdb_handler.rb
lortza, line 104: Instead of raising an error, let's just return out of this method. This will skip the update for this particular movie, but allow the rake task to continue to run for other movies:

```ruby
if movie.title != updated_data.title
  puts "Movie title doesn't match. tmdb_id: #{tmdb_id}. Current title: #{movie.title}. Updated title: #{updated_data.title}"

  return
end
```

Issue #210, opened 2020-07-04
Refactor sorting scopes


mikevallano:
It's very inefficient currently

PR #211, opened 2020-07-04
209: Default sorting to recently added to list

### LINKED ISSUE #209: Default order to recently added to list
Default order to recently added to list

mikevallano:
## Related Issues & PRs
Closes #209

## Problems Solved
* Defaults sort to recently added to the list, with a secondary sort of priority. 

Since we don't really use priority, it's more helpful to order by when a movie was added to the list

## Screenshots
defualt_sort

## Things Learned
* The sorting code is currently a mess. I started refactoring, but opened a separate issue (#210)

### DIFF
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -45,7 +45,7 @@ class Movie < ActiveRecord::Base
   scope :by_recent_release_date, -> { order(:release_date).reverse_order }
   scope :by_highest_vote_average, -> { order(:vote_average).reverse_order }
   scope :default_list_order, -> (list) do
-    list.movies.order('listings.created_at desc, listings.priority desc')
+    list.movies.order('CAST(listings.created_at as DATE) desc, listings.priority desc')
   end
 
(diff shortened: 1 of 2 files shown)


app/models/movie.rb
```diff
@@ -44,6 +44,9 @@ def self.by_tag_and_user(tag, user)
   scope :by_longest_runtime, -> { order(:runtime).reverse_order }
   scope :by_recent_release_date, -> { order(:release_date).reverse_order }
   scope :by_highest_vote_average, -> { order(:vote_average).reverse_order }
+  scope :default_list_order, -> (list) do
+    list.movies.order('listings.created_at desc, listings.priority desc')
+  end
```
lortza, line 49: Just writing this down since we just discussed it...
This secondary sort won't be effective against a timestamp. If we'd like to have this feature work, we'd have to process the `created_at` a little bit, like `listings.created_at.to_date` to make it possible for an entry to have the same date as another, therefore allowing space for the priority ranking.

PR #213, opened 2020-08-18
212: autocomplete list bug

### LINKED ISSUE #212: After adding one movie to a list, subsequent movies don't show lists in the dropdown
Commit that added the autocomplete: https://github.com/mikevallano/tmdb-moviequeue/commit/627743c58e85c714f0a055e8a1924df5ad2c5cf7

list_bug

mikevallano:
## Related Issues & PRs
Closes #212 

## Problems Solved
* Fixes bug with autocomplete on a list view. If you open one modal, then one to the right, it would get fucky.

## Screenshots
| Before      | After |
| ----------- | ----------- |
| list_autocomplete_before  | list_autocomplete_after  |

## Things Learned
The main problem was the JS selectors. 

Jquery autocomplete needs: 1: the selector, and 2: the source to provide the autocomplete options.

We were using generic selectors not specific to the individual movie. So when the initial modal was opened, and autocomplete triggered, it wasn't triggering on modals farther down the DOM.

Similarly, the _source_ needed to be specified. Once I fixed the _trigger_, it would show the wrong lists available after adding the movie to a list.

So, the result is working (again, this whole shebang needs a refactor) with the following:

- In the generic JS file (`z.js`), autocomplete is triggered on a generic selector (`$('.list-dropdown')`). This is for the movie show page because otherwise JS would need to know the movie's `tmdb_id` at that point. Rather than writing the code to do that, I'm just sticking with that generic version.

- After adding a movie to a list, _then_ we use a selector with the movie's `tmdb_id`.  The selector is different in the show page vs the modal.

## Testing
If you want to test it, add a few movies to a list, basically follow what's done in the screenshots, and do some clicky-roundy. I'm confident it's working correctly, but you never know...

### DIFF
--- a/app/views/listings/create.js.erb
+++ b/app/views/listings/create.js.erb
@@ -1,22 +1,21 @@
 var from = "<%= @from %>";
 if (from == 'first_list') {
   console.log("from page " + from);
-  $("#movie_modal_innard_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_modal_innard', :locals => {:movie => @movie}) %>");
-  $("#movie_show_listings_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_show_list_manage', :locals => {:movie => @movie}) %>");
+  $("#movie_modal_innard_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_modal_innard', locals: {movie: @movie}) %>");
+  $("#movie_show_listings_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_show_list_manage', locals: {movie: @movie}) %>");
 }
 else {
-  $("#movie_on_lists_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_on_lists', :locals => {:movie => @movie}) %>");
-  $("#movie_other_list_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_add_to_another_list', :locals => {:movie => @movie}) %>");
-  $("#movie_show_listings_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_show_list_manage', :locals => {:movie => @movie}) %>");
+  $("#movie_on_lists_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_on_lists', locals: {movie: @movie}) %>");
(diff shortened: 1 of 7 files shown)


app/views/listings/create.js.erb
```diff
@@ -1,21 +1,39 @@
...
+  $("#movie_modal_innard_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_modal_innard', locals: {movie: @movie}) %>");
+  $("#movie_show_listings_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_show_list_manage', locals: {movie: @movie}) %>");
 }
 else {
-  $("#movie_on_lists_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_on_lists', :locals => {:movie => @movie}) %>");
-  $("#movie_other_list_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_add_to_another_list', :locals => {:movie => @movie}) %>");
-  $("#movie_show_listings_<%= @movie.tmdb_id %>").html("<%= j(render :partial => 'movies/movie_show_list_manage', :locals => {:movie => @movie}) %>");
+  $("#movie_on_lists_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_on_lists', locals: {movie: @movie}) %>");
+  $("#movie_other_list_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_add_to_another_list', locals: {movie: @movie}) %>");
+  $("#movie_show_listings_<%= @movie.tmdb_id %>").html("<%= j(render partial: 'movies/movie_show_list_manage', locals: {movie: @movie}) %>");
```
lortza, line 10: Nice Scout work on converting those hashrockets. 🚀

app/views/listings/create.js.erb
```diff
@@ -1,21 +1,39 @@
...
 }
 
-$('#add-to-list-dropdown').click(showAllListsForAutocomplete)
+<%# On a list view, we need to specify the movie
```
lortza, line 13: huh! i don't think i've ever seen an erb text comment like this. it totally works though. TIL.

mikevallano: 😆 Yeah, I just did `cmd + /` and that's what happened

app/views/listings/create.js.erb
mikevallano: This file is now 3% less janky 🤣

PR #217, opened 2020-11-11
214: Fix Youtube URL parsing

### LINKED ISSUE #214: Fix trailer urls
When pasting the whole URL from youtube, if other get params are included, such as `ab_channel`, it doesn't work.

So parse to only include `v=`.

For example: https://www.youtube.com/watch?v=y1SpWZm1PLc&ab_channel=A24

mikevallano:
## Related Issues & PRs
Closes #214

## Problems Solved
* Now you can paste youtube urls with whatever additional url params are there
* FULLSCREEN MODE now works! 🎉 

## Screenshots
Fullscreen mode and licecap don't work together, but it works!
trailer_after

## Things Learned
* I modified the code from this SO post: https://stackoverflow.com/a/6557013
* We should probably put some sort of validation in place to ensure nobody can enter invalid stuffs. Since we don't currently have it, I'll open a separate issue for it.

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -78,11 +78,16 @@ class MoviesController < ApplicationController
     end
   end #set movie
 
+  private
+
   def required_params
-    if params[:trailer].present?
-      params[:trailer] = params[:trailer].gsub('https://www.youtube.com/watch?v=', '')
-    end
+    params[:trailer] = parse_youtube_url(params[:trailer]) if params[:trailer].present?
     params.permit(:trailer)
   end
 
+  def parse_youtube_url(trailer)
+    return trailer unless trailer.include?('youtube.com')
+    uri = Addressable::URI.parse(trailer)
+    uri.query_values["v"] if uri.query_values
+  end
 end
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -81,13 +81,13 @@ class MoviesController < ApplicationController
   private
 
   def required_params
-    params[:trailer] = parse_youtube_url(params[:trailer]) if params[:trailer].present?
+    trailer_url = params[:trailer]
+    params[:trailer] = trailer_url.include?('youtube.com') ? youtube_id : trailer_url
     params.permit(:trailer)
   end
 
-  def parse_youtube_url(trailer)
-    return trailer unless trailer.include?('youtube.com')
-    uri = Addressable::URI.parse(trailer)
-    uri.query_values["v"] if uri.query_values
+  def youtube_id
+    uri = Addressable::URI.parse(params[:trailer])
+    uri.query_values['v'] if uri.query_values
   end
 end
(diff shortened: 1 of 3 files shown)


spec/controllers/movies_controller_spec.rb
```diff
@@ -36,19 +36,26 @@
...
       end
 
+      it 'strips youtube url with additional params' do
+        patch :update,
+              { id: movie.id,
+                tmdb_id: movie.tmdb_id,
+                trailer: "https://www.youtube.com/watch?v=#{youtube_id}&ab_channel=A24" }
+        expect(movie.reload.trailer).to eq(youtube_id)
+      end
+
```
lortza, line 53: We may want to add a couple of more cases in here to cover some of the logic in these private methods:
```suggestion
it 'leaves non-youtube urls as-is' do
        original_trailer = 'https://www.example.com'
        patch :update,
              { id: movie.id,
                tmdb_id: movie.tmdb_id,
                trailer: original_trailer }
        expect(movie.reload.trailer).to eq(original_trailer)
      end

      it 'handles empty trailer urls' do
        trailer = ''
        patch :update,
              { id: movie.id,
                tmdb_id: movie.tmdb_id,
                trailer: trailer }
        expect(movie.reload.trailer).to eq(trailer)
      end
```

app/controllers/movies_controller.rb
```diff
@@ -78,11 +78,16 @@ def set_movie
...
-    end
+    params[:trailer] = parse_youtube_url(params[:trailer]) if params[:trailer].present?
     params.permit(:trailer)
   end
 
+  def parse_youtube_url(trailer)
+    return trailer unless trailer.include?('youtube.com')
+    uri = Addressable::URI.parse(trailer)
+    uri.query_values["v"] if uri.query_values
+  end
```
lortza, line 92: When I was reading this, I was a little confused about what was being returned. In one case, we're returning a whole url, in another, we're returning just a youtube_id, but it took me a beat to get there. When I rearrange the two methods like this, it's a little bit more clear _to me_. Of course, it's your call.

```ruby
  def required_params
    trailer_url = params[:trailer]
    params[:trailer] = trailer_url.include?('youtube.com') ? youtube_id : trailer_url
    params.permit(:trailer)
  end

  def youtube_id
    uri = Addressable::URI.parse(params[:trailer])
    uri.query_values['v'] if uri.query_values
  end

```

mikevallano: 👍 Ooh, me likey. Much better!

spec/controllers/movies_controller_spec.rb
mikevallano: 🤔 I did add those, but it's a bit weird because that's not how we _want_ it to work. 

But since it is how it works, I agree it'd be helpful to have the tests there. 

I put them in a context and added a comment referencing the other issue that aims to fix it.

lortza (approved):

PR #220, opened 2020-11-17
218: Restrict trailer edit to admin

### LINKED ISSUE #218: Validate trailer urls
Currently, if someone enters any string, it'll save correctly.

It's a bit tricky because the API sends only the youtube id, not the full URL, and our error handling isn't fully functioning, but we should probably handle it.

mikevallano:
## Related Issues & PRs
Closes #218

## Problems Solved
* Only allow admins to edit movie data. Prevents randos from adding crap data.

## Screenshots
**Non-admins view without the trailer link**
non-admin-after

**This wouldn't happen, but shows the back end validation**
trailer_admin_after

## Things Learned
* Front end tests are still borked 🤷

### DIFF
--- a/spec/controllers/movies_controller_spec.rb
+++ b/spec/controllers/movies_controller_spec.rb
@@ -3,6 +3,7 @@ require 'rails_helper'
 RSpec.describe MoviesController, type: :controller do
 
   let(:user) { create(:user) }
+  let(:admin_user) { create(:user, username: "anne") }
   let(:current_user) { login_with user }
   let(:invalid_user) { login_with nil }
   let(:movie) { create(:movie) }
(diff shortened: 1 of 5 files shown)


spec/controllers/movies_controller_spec.rb
```diff
@@ -27,63 +28,80 @@
...
+        it 'does not update the movie' do
+          @request.env['HTTP_REFERER'] = movie_path(movie)
+          patch :update,
+                  { id: movie.id,
+                    tmdb_id: movie.tmdb_id,
+                    trailer: youtube_id }
+            expect(movie.reload.trailer).not_to eq(youtube_id)
+            expect(response).to redirect_to(movie_path(movie))
+            expect(flash[:alert]).to eq('Must be an admin to access that feature')
+        end
```
mikevallano, line 102: The only change was adding this test. The other diff stuff is just because I put the other tests in a context with an admin user.

spec/features/movies_feature_spec.rb
```diff
@@ -80,7 +86,7 @@
         VCR.use_cassette('update_movie') do
           click_link("update_movie_link_movie_show")
         end
-        expect(page).to have_content("98 min")
+        expect(page).to have_content("1h 38m")
```
mikevallano, line 89: Unrelated. Just noticed it was failing, and we apparently never updated it when we changed the minute count display.

lortza (approved):
Looks good to me. 👍  
Separately, I'd love to move our admin logic into a boolean here in the near future.

PR #224, opened 2020-11-20
223: Add admin boolean

### LINKED ISSUE #223: De-jankify admin logic
Add boolean column

mikevallano:
## Related Issues & PRs
Closes #223

## Problems Solved
* Adds a proper boolean field to determine admins.

## Screenshots
N/A

## Things Learned
* We still don't have a way to _assign_ admin, but that's not going to be a problem for the foreseeable future. After the deploy, I'll update the records directly in the Rails console. I thought about doing it in the migration, but it seemed a bit insecure to add that in the codebase 🤷

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,5 @@ before_script:
   - psql -c 'create database travis_ci_test;' -U postgres
   - export DISPLAY=:99.0
 script:
-  - bundle exec rake db:setup
+  - bundle exec rake db:migrate RAILS_ENV=test
   - bundle exec rspec --exclude-pattern "**/features/*_spec.rb"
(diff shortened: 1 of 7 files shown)


lortza (approved):
This looks good to me. 👍  We may discover that we have more of this logic nested through a bunch of code, but i'm comfortable with deploying this now and then fixing bugs as they come up.

mikevallano:
I'm sorry but I don't have 🐛 in my code 😆 .

lortza:
Of _course_, of _course_. My mistake.

lortza:
I noticed the schema file wasn't part of this PR. Might that be why Travis was having problems?

PR #229, opened 2021-03-05
Add link to movie poster view

### LINKED ISSUE #228: Improve actor credit/movies displays. Include link back and forth between the two views
We have two different views: one that shows bulleted list of credits, and the other that shows the movie poster.

But we don't have a way to go between the two. 

Think through what we want and make it happen.

lortza:
## Related Issues & PRs
Closes #228

## Problems Solved
We wanted a simple link to get us between the actor bio and the view where we see the posters of all of the movies they were in. This adds a link to the top of the "movie credits" section of the actor bio page that lets us get to that movie poster view. It's simple, but it works. I was considering putting something prettier and more stand-out-looking on the page, but I wasn't sure where to put it where it wouldn't be awkward. Any ideas?

## Screenshots
Screen Shot 2021-03-04 at 7 03 37 PM

### DIFF
--- a/app/helpers/movie_person_profiles_helper.rb
+++ b/app/helpers/movie_person_profiles_helper.rb
@@ -7,4 +7,8 @@ module MoviePersonProfilesHelper
     date = profile.birthday.to_date.stamp('January 1st, 2018')
     profile.deathday.blank? ? "#{date} (Age: #{profile.age})" : "#{date}, Deceased (Age: #{profile.age})"
   end
+
+  def actor_movie_posters_uri(actor)
+    "#{root_url}/tmdb/actor_search?actor=#{actor.name.gsub(' ', '+')}"
+  end
 end
--- a/app/helpers/movie_person_profiles_helper.rb
+++ b/app/helpers/movie_person_profiles_helper.rb
@@ -9,6 +9,6 @@ module MoviePersonProfilesHelper
   end
 
   def actor_movie_posters_uri(actor)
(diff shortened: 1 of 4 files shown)


app/helpers/movie_person_profiles_helper.rb
```diff
@@ -7,4 +7,8 @@ def display_birthday_info(profile)
     date = profile.birthday.to_date.stamp('January 1st, 2018')
     profile.deathday.blank? ? "#{date} (Age: #{profile.age})" : "#{date}, Deceased (Age: #{profile.age})"
   end
+
+  def actor_movie_posters_uri(actor)
```
mikevallano, line 11: Was reading the [difference between](https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn) `uri` and `url` and 🤷 . I don't care, and you don't need to explain it here, but just curious.

mikevallano (approved):
👍 Looks good where it is. Grassy! 🤗

app/helpers/movie_person_profiles_helper.rb
lortza: Though it bothered me, I made that decision for 3 reasons:
* I didn't want any confusion between that helper method and a rails-generated  path helper -- for the sake of troubleshooting.
* I knew it wasn't _wrong_ even though it didn't feel like the right name overall
* naming is extra hard when you're extra tired 😆 😪

lortza:
How do you get coveralls to rerun a build?

PR #231, opened 2021-04-27
Remove awkward gap between image and buttons in modal on tablet

lortza:
## Problems Solved
There is an awkward gap between image and buttons in modal on tablet. This PR resolves it.

## Screenshots
Before
Screen Shot 2021-04-27 at 1 18 48 PM

After
Screen Shot 2021-04-27 at 1 12 05 PM
Screen Shot 2021-04-27 at 1 12 18 PM

## Things Learned
* We have some sections that identify themselves with bootstrap classes and we should be naming things better.

### DIFF
--- a/app/views/movies/_movie.html.erb
+++ b/app/views/movies/_movie.html.erb
@@ -12,9 +12,7 @@
       <div class="modal-body">
         <div class="row">
         <div class="col-xs-3">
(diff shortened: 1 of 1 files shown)


mikevallano (approved):
So much better! We decided to move the `x` is a separate issue.

PR #233, opened 2021-05-16
Add environment banner

lortza:
## Problems Solved
* Makes is clear when you are on localhost by adding a banner to the top of the page
* The banner stays fixed to to the top when you scroll
* Includes branch name, which is useful to keep you from working on master
* Only appears in development or test environments

## Screenshots
Screen Shot 2021-05-16 at 2 56 44 PM

## About the Work
* The approach I took with this code was to encapsulate it as much as possible because I'd like to move it out into a gem. For this reason, you'll see inline styles and a lib file instead of a typical `app/helpers` file and changes to the stylesheets.
* I did have to adjust our navbar a little bit and part of how we set up the `layouts/application.html.erb` file, but I was fine with that because it was a little bit janky and I got to fix that.

### DIFF
--- a/app/assets/stylesheets/layouts.scss
+++ b/app/assets/stylesheets/layouts.scss
@@ -2,14 +2,7 @@
 @import "fonts";
 
 
-.row {
-  margin: 10px 0px;
-  // clear: both;
-} /* .row */
-
-.shover {
-  height: 50px;
-} /* .shover */
+.row { margin: 10px 0px; }
 
 .navbar-brand {
   font-family: $header-family;
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-default navbar-fixed-top navbar-inverse">
+<nav class="navbar navbar-default navbar-inverse">
   <div class="container-fluid">
     <!-- Brand and toggle get grouped for better mobile display -->
     <div class="navbar-header">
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/layouts.scss
```diff
@@ -2,14 +2,7 @@
 @import "fonts";
 
 
-.row {
-  margin: 10px 0px;
-  // clear: both;
-} /* .row */
-
-.shover {
```
lortza, line 10: we used to have a "shover" div on the application layout because the navbar was fixed to the top. We don't need it anymore.

app/views/layouts/_header.html.erb
```diff
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-default navbar-fixed-top navbar-inverse">
```
lortza, line 1: I removed the `fixed` nature of this navbar because it was problematic for the env banner. It was not a useful quality and was also causing us some z-index issues.

app/views/layouts/application.html.erb
```diff
@@ -1,37 +1,31 @@
...
+  <body>
+    <%= render "layouts/environment_banner" %>
+    <%= render "layouts/header" %>
+    <div class="container">
+      <p id="notice"><%= notice %></p>
+      <p id="alert"><%= alert %></p>
+      <%= render "shared/overlay_modal" %>
+      <%= yield %>
+    </div>
+  </body>
```
lortza, line 30: The changes to this file are mostly in indention. But in this section, I did rearrange the divs a little bit to make the navbar smarter and allow space for the env banner.

mikevallano (approved):
😎

PR #234, opened 2021-05-16
Add gem better_errors

lortza:
## Problems Solved
The Better Errors gem gives us... better errors 😆 including access to a console where you can check values of any variables available to you at the moment of failure. While the Rails default has some of the features this gem has now, I prefer the better_errors interface. 

## Screenshots
Before
Screen Shot 2021-05-16 at 4 54 00 PM

After
Screen Shot 2021-05-16 at 4 51 35 PM

## About the Work
* this is a development-env-only change
* it does not affect app code
* it just makes dev life nicer

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,8 @@ gem 'sentry-raven'
 
 group :development do
   gem 'web-console', '~> 2.0'
+  gem 'better_errors'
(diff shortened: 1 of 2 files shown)


mikevallano (approved):
👍

PR #250, opened 2021-12-19
Add tv show search

lortza:
## Problems Solved
We don't currently have a way to search for TV shows on our app. Granted, we built the app to be a MovieQueue™️ 😆 and not a TV tracking app. However, when we're watching some of our favorite TV shows, I find myself going to IMDB to look up actors. I prefer the data and interface of our own app, so I added a simple TV search to get us to the TV Series pages that already exist in our app. 

## Screenshots
The old (current) workflow requires knowing at least one actor's name (tough, considering not may of these folks were famous), searching by their name, going to their bio, scrolling down to their credit for the show, clicking on the show there.
Kapture 2021-12-19 at 14 34 53

The new way allows me to search for the show directly
Kapture 2021-12-19 at 14 25 55

## Things Learned
* I looked into the results we get for the "full cast" search because you and I were talking about another feature where we'd add an actor's current age to their listing. Birth date is not returned in that response. We'd have to do another call per actor to get that info for that page. 
* As we discussed, having instance variables buried way down in service classes is gnarly for adding features or refactoring. I'm glad to struggle with it today though. It gave me a very clear example (instead of just theory) as to why this pattern is so painful and should be avoided.
* jQuery still flummoxes me. But 🪄 I was able to fix it. 🎉

### DIFF
--- a/app/assets/javascripts/z.js
+++ b/app/assets/javascripts/z.js
@@ -8,14 +8,18 @@ $(document).ready(function(){
       $("#overlay-modal").show(); }
     });
 
+  $("#autocomplete-tv-series-search-submit").autocomplete({
+    source: $("#autocomplete-tv-series-search-submit").data("autocomplete-source"),
+    minLength: 4,
+    select: function(event, ui) {
+      $("#autocomplete-tv-series-search-submit").val(ui.item.value);
+      $(this).closest("form").submit();
+      $("#overlay-modal").show(); }
+    });
+
   $( ".autocomplete-search-field" ).autocomplete({
     minLength: 4,
     source: $(".autocomplete-search-field").data("autocomplete-source")
-  });
-
-  $( ".autocomplete-tv-series-search" ).autocomplete({
-    minLength: 4,
-    source: $(".autocomplete-tv-series-search").data("autocomplete-source")
   });
 
   $('#header-movie-search, ul.ui-autocomplete').css({
--- a/app/assets/javascripts/z.js
+++ b/app/assets/javascripts/z.js
@@ -1,21 +1,18 @@
 $(document).ready(function(){
-  $(".autocomplete-auto-submit").autocomplete({
-    source: $(".autocomplete-auto-submit").data("autocomplete-source"),
-    minLength: 4,
-    select: function(event, ui) {
-      $(".autocomplete-auto-submit").val(ui.item.value);
-      $(this).closest("form").submit();
-      $("#overlay-modal").show(); }
-    });
+  $('.autocomplete-auto-submit').each(function(){
+    let form = $(this)
+    let source = form.data('autocomplete-source')
 
-  $("#autocomplete-tv-series-search-submit").autocomplete({
-    source: $("#autocomplete-tv-series-search-submit").data("autocomplete-source"),
-    minLength: 4,
-    select: function(event, ui) {
-      $("#autocomplete-tv-series-search-submit").val(ui.item.value);
-      $(this).closest("form").submit();
-      $("#overlay-modal").show(); }
-    });
+    form.autocomplete({
+      source: source,
+      minLength: 4,
+      select: function (event, ui) {
+        $(this).closest(".autocomplete-auto-submit").val(ui.item.value);
+        $(this).closest("form").submit();
+        $("#overlay-modal").show();
+      }
+    })
+  });
 
   $( ".autocomplete-search-field" ).autocomplete({
     minLength: 4,
--- a/app/assets/javascripts/z.js
+++ b/app/assets/javascripts/z.js
@@ -13,6 +13,11 @@ $(document).ready(function(){
     source: $(".autocomplete-search-field").data("autocomplete-source")
   });
 
+  $( ".autocomplete-tv-series-search" ).autocomplete({
+    minLength: 4,
+    source: $(".autocomplete-tv-series-search").data("autocomplete-source")
+  });
+
   $('#header-movie-search, ul.ui-autocomplete').css({
     'position': 'relative',
     'max-width': '455px'
--- a/app/assets/stylesheets/movies.scss
+++ b/app/assets/stylesheets/movies.scss
@@ -30,7 +30,7 @@ article.tile-cover-pic {
   position: relative;
   background-color: $missing-poster-background-color;
   float: left;
-  margin: -10px 10px 10px 0px;
+  margin: 10px 10px 10px 0px;
   width: $index-poster-width;
   height: $index-poster-height;
   word-wrap: break-word;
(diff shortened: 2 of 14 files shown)


app/assets/stylesheets/movies.scss
```diff
@@ -30,13 +30,14 @@ article.tile-cover-pic {
   position: relative;
   background-color: $missing-poster-background-color;
   float: left;
-  margin: -10px 10px 10px 0px;
+  margin: 10px 10px 10px 0px;
```
lortza, line 33: This css change allows the TV show name text to fit more gracefully on the "missing image" poster.

app/helpers/movies_helper.rb
```diff
@@ -1,13 +1,11 @@
...
-
   def image_for(movie)
     if movie.poster_path.present?
       image_tag("http://image.tmdb.org/t/p/w185#{movie.poster_path}", title: movie.title, alt: movie.title)
     else
-      render "movies/movie_missing_poster", movie: movie
-      # image_tag('placeholder.png')
-    end #if
-  end #image_for
+      render "shared/missing_poster", title: movie.title
```
lortza, line 6: This is such a tiny partial with no logic in it. I decided to move it into the `shared` directory so it can be used for both movies and tv.

app/helpers/tv_series_helper.rb
```diff
@@ -0,0 +1,11 @@
+module TvSeriesHelper
+
+  def image_for_tv_poster(series)
```
lortza, line 3: I could not easily reuse the `image_for` helper that we made for movies without a larger refactor. I decided to make a new version for TV instead.

app/models/tv_series.rb
```diff
@@ -1,6 +1,5 @@
 class TVSeries
-  def initialize(show_id, first_air_date, last_air_date, show_name, backdrop_path, poster_path, number_of_episodes, number_of_seasons, overview, seasons, actors)
-
+  def initialize(show_id:, first_air_date:, last_air_date:, show_name:, backdrop_path:, poster_path:, number_of_episodes:, number_of_seasons:, overview:, seasons:, actors:)
```
lortza, line 2: I converted this to keyword arguments because there are so many arguments, it was difficult to keep them straight.

app/models/tv_series.rb
```diff
@@ -29,7 +46,19 @@ def self.parse_results(result, show_id)
     @overview = result[:overview]
     @seasons = (1..(result[:number_of_seasons])).to_a if result[:number_of_seasons].present?
     @actors = TVCast.parse_results(result[:credits][:cast])
-    @series = TVSeries.new(@show_id, @first_air_date, @last_air_date, @show_name, @backdrop_path, @poster_path, @number_of_episodes, @number_of_seasons, @overview, @seasons, @actors)
-  end #parse results
 
-end #class
\ No newline at end of file
+    @series = TVSeries.new(
```
lortza, line 50: I needed to change this line because of the keyword argument refactor above. I'd have preferred to roll lines 38-48 above in to these lines below, however, I am not sure where all of the instance variables are being called further down the line. I can do a separate cleanup PR to address this. In the meantime, I chose to keep the scope of this PR as small as possible.

app/views/shared/_missing_poster.erb
```diff
@@ -0,0 +1,3 @@
+<div class="missing-poster">
+	<p><%= title %></p>
```
lortza, line 2: This partial used to take a `movie` object and call `movie.title`. Now it just takes the text we want to render (`title`). This means the partial can be used for both movies (that have a `title` attribute) and TV shows (that have a `show_name` attribute).

app/views/tmdb/tv_series_search.html.erb
```diff
@@ -0,0 +1,28 @@
+<% content_for(:title, "Search by TV Show Title") %>
```
lortza, line 1: This is the new search page that is the center of this PRs work.

lib/tmdb_handler.rb
```diff
@@ -152,6 +152,19 @@ def tmdb_handler_actor_credit(credit_id)
     @credit = TVActorCredit.parse_results(@credit_results)
   end
 
+  def tmdb_handler_tv_series_search(query)
+    search_url = "#{BASE_URL}/search/tv?query=#{query}&api_key=#{ENV['tmdb_api_key']}"
```
lortza, line 156: This is the exciting new `/search/tv` endpoint. I learned today that they also have a `/search/multi` endpoint, which is neat, but incorporating it into our current workflows would get messy quickly.

app/assets/javascripts/z.js
```diff
@@ -8,6 +8,15 @@ $(document).ready(function(){
...
 
+  $("#autocomplete-tv-series-search-submit").autocomplete({
+    source: $("#autocomplete-tv-series-search-submit").data("autocomplete-source"),
+    minLength: 4,
+    select: function(event, ui) {
+      $("#autocomplete-tv-series-search-submit").val(ui.item.value);
+      $(this).closest("form").submit();
+      $("#overlay-modal").show(); }
+    });
+
```
mikevallano, line 19: 😓 I was hacking various things trying to get this to work without adding a new function, and the following hack (I mean, everything in jquery feels like a hack) seems to work:

```js
  $('.autocomplete-auto-submit').each(function(){
    let form = $(this)
    let source = form.data('autocomplete-source')

    form.autocomplete({
      source: source,
      minLength: 4,
      select: function (event, ui) {
        $(".autocomplete-auto-submit").val(ui.item.value);
        $(this).closest("form").submit();
        $("#overlay-modal").show();
      }
    })
  });
```

app/models/tv_series.rb
mikevallano: 🎉

app/views/shared/_missing_poster.erb
mikevallano: ❤️

mikevallano (commented):
😍 Sweet featch! Nice refactoring, but not _too much_ refactoring too!

I just had the one comment about tweaking the autocomplete jquery to be flexible rather than adding a new function for the one form, but other than that it's 💯

app/assets/javascripts/z.js
lortza: 🎉 😮 Nice! This totally works for both movies and tv series. There is a brief moment right after you select the item from the dropdown and before the submit action kicks in that the selected title is populated in both of these search fields. 
Screen Shot 2021-12-20 at 7 29 32 PM
How do you feel about that functionality vs duplicating a function in js?

lortza: Actually, this fixes it:
```javascript
// replace 
$(".autocomplete-auto-submit").val(ui.item.value);

// with this
$(this).closest(".autocomplete-auto-submit").val(ui.item.value);
```

mikevallano (approved):
🎉

mikevallano:
Deployed! FYI, I had to change `let` to `var` because the deploy failed due to old school JS setup.

PR #251, opened 2021-12-19
Refactor: reduce and centralize instance variables in TVSeries methods

lortza:
## Related Issues & PRs
* Follow up to PR https://github.com/mikevallano/tmdb-moviequeue/pull/250
* Precursor to PR #252

## Problems Solved
I discovered while refactoring the `TVSeries` object in #250 that it needed a little bit more work to safely remove all of the instance variables. To keep the scope of that PR focused, I pulled that continued refactoring work out into this PR. 

We've learned that it's much more sustainable maintenance-wise to have our instance variables be as few and as close to the surface as possible. This PR and PR #252 continue the work of pulling instance variables out of the `tmdb_handler` service object methods and centralizing their values under fewer instance variables (ideally a single instance variable) in the corresponding `tmdb_controller`.  

This PR does not change any functionality. It is refactor & cleanup only.

PR #252, opened 2021-12-21
Refactor: separating concerns with TVEpisode and TVSeason

lortza:
## Related Issues & PRs
* Follow up to PR #251

## Problems Solved
While working on #251, a line caught my eye ` @episode = TVSeason.new(...` so it tipped me off that we had a little blurring of concerns with these two objects. The work in this PR is focused on 2 main goals
1. reducing and centralizing instance variables within the `tmdb_controller`
2. drawing a clear line between the purpose of the `TVEpisode` and `TVSeason` objects

This PR does not change any functionality. It is refactor & cleanup only.

## Things Learned
It can be tricky to decide how to design an object. Do we `new` it up with all of its attributes or maybe just a data object to parse? It depends on how you want to use this object and where you think you'll be instantiating it in the future. I chose to instantiate with with its attributes because this gives us more flexibility for where and when we can instantiate another one in other parts of the app.


app/controllers/tmdb_controller.rb
```diff
@@ -110,9 +110,13 @@ def tv_series
 
   def tv_season
     show_id = params[:show_id]
-    @season_number = params[:season_number]
+    season_number = params[:season_number]
     @series = tmdb_handler_tv_series(show_id)
-    @episodes = tmdb_handler_tv_season(show_id, @season_number)
+    @season = tmdb_handler_tv_season(
```
lortza, line 115: This change puts the `episodes` inside the `season` object, which is the main object of the page.

app/models/tv_actor_credit.rb
```diff
@@ -1,35 +1,32 @@
...
-    if result[:media][:seasons].present?
-      @seasons = TVCreditSeasons.parse_results(result[:media][:seasons])
-    else
-      @seasons = nil
-    end
-    if result[:media][:episodes].present?
-      @episodes = TVEpisodes.parse_results(result[:media][:episodes])
-    else
-      @episodes = nil
-    end
```
lortza, line 31: Here, I'm taking advantage of rails' `presence` to safely return `nil`. We no longer need the if/else to do that for us.

I've also renamed `TVCreditSeasons --> TVCreditSeason` and `TVEpisodes --> TVEpisode` to reflect the singular nature of these objects.

app/models/tv_actor_credit.rb
```diff
@@ -1,35 +1,32 @@
...
-  end #init
-
-  attr_accessor :show_name, :show_id, :actor_name, :actor_id, :character, :seasons, :episodes
-
-  def self.parse_results(result)
-    @show_name = result[:media][:name]
-    @show_id = result[:media][:id]
-    @actor_name = result[:person][:name]
-    @actor_id = result[:person][:id]
-    @character = result[:media][:character]
```
lortza, line 21: I've reconciled and removed these instance variables. They're accessible only as attributes in this object now.

app/models/tv_credit_season.rb
```diff
@@ -0,0 +1,34 @@
...
+    @show_id = show_id
+  end
+
+  def self.parse_records(json)
+    json.map do |record|
+      parse_record(record)
+    end.sort_by { |r| r.season_number }
+  end
+
+  def self.parse_record(record)
```
lortza, line 21: I've broken the parsing into 2 different methods. This allows the caller to choose whether they're trying to parse a single object or an array of these objects. I chose the word `record` because it is easy to tell when it is single/plural.

app/models/tv_episode.rb
```diff
@@ -0,0 +1,35 @@
+class TVEpisode
```
lortza, line 1: This file has the same pattern or refactoring:
* rename to singular version
* remove instance variables
* create `parse_record` and `parse_records` methods

app/models/tv_season.rb
```diff
@@ -1,34 +1,32 @@
...
-    @episode_number = episode_number
-    @name = name
+  attr_accessor :series, :show_id, :air_date, :name, :overview, :season_id, :poster_path, :season_number, :credit, :episodes
+
+  def initialize(series:, show_id:, air_date:, name:, overview:, season_id:, poster_path:, season_number:, credit:, episodes:)
+    @series = series
+    @show_id = show_id
     @air_date = air_date
-    @guest_stars = guest_stars
+    @name = name
```
lortza, line 8: This class was having a little bit of an identity crisis. It was mixing the ideas of `season` and `episode`.  This refactor separates those concerns while still delivering the goods to the html pages.

app/views/tmdb/actor_credit.html.erb
```diff
@@ -1,52 +1,43 @@
 <article id="profile">
```
lortza, line 1: While this looks like a lot of changes here, they're really whitespace changes. I cleaned up some rogue closing tags  and adjusted the indenting.

app/views/tmdb/tv_season.html.erb
```diff
@@ -24,10 +24,10 @@
 </div> <!-- row -->
 
 <div class="row">
-  <h2><%= formatted_season_number(@season_number) %> Episodes</h2>
+  <h2><%= @season.name %> Episodes</h2>
   <table class="table">
     <tbody>
-      <% @episodes.each do |episode| %>
+      <% @season.episodes.each do |episode| %>
```
lortza, line 30: `episodes` lives inside of the `season` object now.

mikevallano (approved):
👏 🎉 🚀

Issue #249, opened 2021-12-12
Add actor age-at-time to Full Cast page


lortza:
I frequently find myself wondering how old an actor was when a movie was made. Adding a little math to their listing on the full cast page should be  an easy lift. It may require a light redesign. I'm game for that.

mikevallano:
You checked and the birthdate is not included in the cast response, but we should check to see if we can send one request with multiple actor ids rather than one request per actor

lortza:
I think I have a compromise that can move us forward for the moment. On the actor bio page there are movie credits that show a year and we know the actors birth year. We can do the math and show it in the movie credit.

lortza:
I was not able to find any sort of group id endpoint for actor bios. :( So I'm not really sure what our next move would be here... 🤔 assuming we're not trying to persist any actor information in our own db.

mikevallano:
Yeah, bummer. Maybe we could make a feature request. Seems like it should be relatively easy for them to add one more piece of data (`birthday`) to the https://developers.themoviedb.org/3/movies/get-movie-credits response. 

Looks like the feature request could be made in the forum:
https://www.themoviedb.org/talk/category/5047958519c29526b50017d6

PR #255, opened 2021-12-24
Add actor age to movies credits on bio page

lortza:
## Related Issues & PRs
* Partially addresses Issue https://github.com/mikevallano/tmdb-moviequeue/issues/249

## Problems Solved
* I'm always curious how old an actor is when they're in a movie
* I don't like doing math
* Only displays if there is enough information to do the math
* I'd like this info to be on the movie cast page, but we've got some n+1 issues. This is a compromise.

## Screenshots
Before: how in the heck old is Tommy Lee Jones in these movies??
Screen Shot 2021-12-24 at 10 48 06 AM

After: oh, okay. Has he _always_ been in his 70s??
Screen Shot 2021-12-24 at 10 49 39 AM

No, actually, he was in his 40s once. But how old was he in the MIBs?
Screen Shot 2021-12-24 at 10 52 59 AM

Oh, okay.
Screen Shot 2021-12-24 at 10 53 47 AM

### DIFF
--- a/app/helpers/movies_helper.rb
+++ b/app/helpers/movies_helper.rb
@@ -7,6 +7,11 @@ module MoviesHelper
     end
   end
 
+  def actor_age_at_movie_release(actor_birthday, release_year)
+    birth_year = actor_birthday.to_date.year
+    release_year.to_i - birth_year
+  end
+
   def link_to_movie(movie)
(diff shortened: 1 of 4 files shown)


mikevallano (approved):
😎

PR #259, opened 2021-12-30
257: Remove find similar movies feature

### LINKED ISSUE #257: Look into "find similar movies"
@mikevallano this is a note for you

I dobelive you want to scrap this feature because it is ineffective and we don't use it.

mikevallano:
## Related Issues & PRs
Closes #257

There's a big diff, but most of it is VCR cassettes because we had been requesting similar movies in requests where we weren't using it, so since the request changed, the VCR cassettes needed to be updated.

## Problems Solved
* The results provided by TMDB for similar movies isn't useful, so we never use this feature, and this removes the code.

## Screenshots
**Before**
similar-before

**After**
similar-after

## Things Learned
* We had a chunk of code for a feature we didn't use ☹️ 
* I created a [separate issue](https://github.com/mikevallano/tmdb-moviequeue/issues/266) to make sure we're using `append_to_response` only when needed


lortza (approved):
💯 Looks great! I love that you're killing off a feature that we never use and cleaning out the codebase. 🙌 💃

PR #261, opened 2021-12-30
Add images for season posters

lortza:
## Related Issues & PRs
Closes #X

## Problems Solved
*

## Screenshots
[screenshots here]

## Things Learned
*

PR #262, opened 2021-12-30
Replace individually built image urls with a helper method

lortza:
## Problems Solved
There are several hard-coded urls for the TMDB image domain in our views. While exploring images for TV seasons and episodes, I found myself going down a small API doc rabbit hole in PR #261, trying to find the magic number required to render images for "stills", "backdrops", "posters", "profiles", etc, as they all have different sizes available. Yikes.

The [image docs](https://developers.themoviedb.org/3/getting-started/images) send you to the [configuration docs](https://developers.themoviedb.org/3/configuration/get-api-configuration) where they recommend hitting this endpoint to get up-to-date images sizes. More yikes. Since 2016, TMDB has not changed the image sizes, so I don't see a need for us to implement any caching for this data anytime soon. A static file is fine until it becomes problematic.

I decided to build a module to house this information. My goals are:
* We were not using the secure image endpoint. Now we are.
* When working on this app, we don't need to know exact pixel size most of the time. Choosing from a set list of `xxsmall` to `xxlarge` options for sizes makes things feel easy.
* Not every image type has the same available sizes. I've included error messages to help the developer when they use this method. Additionally, just going to the file and looking at the available sizes is easy peasy compared to digging around in the api endpoint.
* Most of the time, we're displaying "poster" images at size "medium" setting those values as defaults saves us keystrokes.
* if TMDB changes the pixel size for a "medium" "poster", we can update it here in the helper and then _not_ have to update it in the 14 other files where it is being used.

## Things Learned
* Naming is hard... sometimes. This was one of those times. 
* Remember that anything in the rails `Helpers` dir is loaded globally. Cool and _dangerous_. Since my new service method is called `image_url` that is in direct conflict with the rails method of the same name. My method now lives in the `lib` folder and can only be invoked directly on its module name: `TmdbImageService.image_url`. Inside the helpers directory, it could have been called by `image_url` and been in a race condition with rails. I don't like racing with rails because I know who's going to lose. [spoiler: me]

### DIFF
--- a/app/helpers/movies_helper.rb
+++ b/app/helpers/movies_helper.rb
@@ -1,7 +1,13 @@
+# frozen_string_literal: true
+
 module MoviesHelper
   def image_for(movie)
     if movie.poster_path.present?
-      image_tag("http://image.tmdb.org/t/p/w185#{movie.poster_path}", title: movie.title, alt: movie.title)
+      image_tag(
+        TmdbImageUrlHelper.image_url(file_path: movie.poster_path, size: :medium, image_type: :poster),
+        title: movie.title,
+        alt: movie.title
+      )
     else
       render "shared/missing_poster", title: movie.title
     end
--- a/app/helpers/movies_helper.rb
+++ b/app/helpers/movies_helper.rb
@@ -4,7 +4,7 @@ module MoviesHelper
   def image_for(movie)
     if movie.poster_path.present?
       image_tag(
-        TmdbImageUrlHelper.image_url(file_path: movie.poster_path, size: :medium, image_type: :poster),
+        TmdbImageService.image_url(file_path: movie.poster_path, size: :medium, image_type: :poster),
         title: movie.title,
         alt: movie.title
       )
--- a/lib/tmdb_image_service.rb
+++ b/lib/tmdb_image_service.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module TmdbImageUrlHelper
+module TmdbImageService
   # Image Docs: https://developers.themoviedb.org/3/getting-started/images
   # Image configuration: https://developers.themoviedb.org/3/configuration/get-api-configuration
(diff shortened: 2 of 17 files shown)


spec/features/movies_feature_spec.rb
```diff
@@ -53,7 +53,7 @@
         sign_in_user(user)
         listing
         visit(movie_path(movie))
-        expect(page).to have_css("img[src*='http://image.tmdb.org/t/p/w185/aZeX4XNSqa08TdMHRB1gDLO6GOi.jpg']")
+        expect(page).to have_css("img[src*='https://image.tmdb.org/t/p/w185/aZeX4XNSqa08TdMHRB1gDLO6GOi.jpg']")
```
lortza, line 56: This is part of the switch from `http` to `https`

spec/lib/tmdb_image_service_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe TmdbImageService do
```
lortza, line 5: These specs test all of the logic that is added with this new service method.

lib/tmdb_image_service.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module TmdbImageService
```
lortza, line 3: This is our exciting new service that houses the api `/configuration` endpoint data and the url helper method.

app/helpers/movies_helper.rb
```diff
@@ -1,7 +1,13 @@
+# frozen_string_literal: true
```
lortza, line 1: The bulk of the changes in the next 10 files will be like the one in this file: changing out the hardcoded url with the new service method.

mikevallano (approved):
🎉 😎

PR #263, opened 2021-12-30
Resolve bug where age is not able to be calculated

### LINKED ISSUE #258: Calculating age at release errors when actor birthday isn't present
To reproduce, try `tmdb/actor_more?actor_id=205560`

age_at_release

lortza:
## Related Issues & PRs
Closes #258

## Problems Solved
* handles when an actor's age can't be calculated
* displays no age information if we don't have an age
* displays `| age 42` when an actor's age is 42

## Screenshots
Here's our main culprit: the actor without a birth year. Now you can see it doesn't say a peep about age.
Screen Shot 2021-12-30 at 4 00 25 PM

Here's an example where the first line doesn't know about age for some reason, but subsequent lines do:
Screen Shot 2021-12-30 at 4 00 38 PM

## Things Learned
* Conditions are annoying
* I had a choice of putting more conditional logic in the erb or moving it into a new helper method. I chose the latter.

### DIFF
--- a/app/helpers/movies_helper.rb
+++ b/app/helpers/movies_helper.rb
@@ -7,7 +7,14 @@ module MoviesHelper
     end
   end
 
+  def display_actor_age_at_release(actor_birthday, release_year)
+    age = actor_age_at_movie_release(actor_birthday, release_year)
+    "| age #{age}" if age.present?
+  end
+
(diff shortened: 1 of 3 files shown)


mikevallano (approved):
🥇

PR #264, opened 2021-12-31
Rename tmdb_handler_ methods as class-level

lortza:
Blocked by #270

## Problems Solved
This is a cleanup PR. It is a simple rename and does not change any functionality. We decided to rename all of the methods in the `TmdbHandler` module to make them more idiomatic by removing `tmdb_handler_` and prepending `self`. We prefer to see the `self.` notation over the `self << class` notation and since we don't have any private methods in this module, it works for now. It's not 100% squeaky clean in scope as I did take some liberties in cleaning up some method-end comments and convert a couple of `if` statements into one-liners.

This sets us up to be able to refactor this module in subsequent PRs, including some things we talked about:
* moving this to a namespace: `Tmdb::Handler` along with `Tmdb::ImageService`
* maybe renaming this to something like `Tmdb::Client` or `Tmdb::Api`
* maybe pulling some logic into private methods

### DIFF
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -34,7 +34,7 @@ class TmdbController < ApplicationController
     if params[:tmdb_id]
       @tmdb_id = params[:tmdb_id]
       movie = Movie.find_by!(tmdb_id: @tmdb_id)
-      TmdbHandler.TmdbHandler.update_movie(movie)
+      TmdbHandler.update_movie(movie)
     end
     redirect_to movie_more_path(tmdb_id: @tmdb_id)
   end
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -1,31 +1,30 @@
 class TmdbController < ApplicationController
-  before_action :authenticate_user!
-
   require 'open-uri'
-  include TmdbHandler
   include SearchParamParser
+
+  before_action :authenticate_user!
 
   def search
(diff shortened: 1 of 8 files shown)


app/controllers/tmdb_controller.rb
```diff
@@ -35,7 +34,7 @@ def update_tmdb_data
     if params[:tmdb_id]
       @tmdb_id = params[:tmdb_id]
       movie = Movie.find_by!(tmdb_id: @tmdb_id)
-      TmdbHandler.tmdb_handler_update_movie(movie)
+      TmdbHandler.TmdbHandler.update_movie(movie)
```
lortza, line 37:
```suggestion
      TmdbHandler.update_movie(movie)
```

lortza:
Something very strange is happening here. We're getting errors because the instance variables set in the `TmdbHandler` are no longer making it to the `tmdb_controller`. They're being set in the handler and I have confirmed they have a value, but they are nil in the controller action. I was able to resolve this for the `director_search` action by cleaning up the instance variables and setting a single variable in the controller -- which is what we'd prefer, but that's a much bigger refactor than the scope of this PR.

I'm not sure why this is happening... 🤔

lortza:
I opened PR https://github.com/mikevallano/tmdb-moviequeue/pull/265 to help get our instance variables cleaned up.

lortza:
We'll, it's been a journey! I'm closing this because I've made significant changes since opening this PR and now I'm taking a different approach via PR #284.

PR #265, opened 2021-12-31
Instance variable refactors for director_search and actor_more

lortza:
## Problems Solved
While working on https://github.com/mikevallano/tmdb-moviequeue/pull/264, I found myself in the swirling twirling land of instance variable monsters. I opened this PR to start chipping away at some of those instance variables that we'd set in the `TmdbHandler` and `tmdb_controller`. Since this work can spiral, my goal was to keep this PR as small as possible. I was able to clean up 2 controller methods, 2 tmdb_handler methods, and a handful of partials. 

I'd like to continue this work in subsequent PRs, trying to keep them as small as possible.

## Risk
These little beasties are sneaky. I traced the code of every time these views are rendered and made sure anyone who renders the partial has the data needed to render it properly. A failure would look like a `NilClass` error. Aside from tests, attentive code tracing, and browser testing, I'm not sure how else to mitigate this risk.

## TODO:
`tmdb_controller` actions:
- [x] `director_search`
  - [x] erb: director_search
    - [x] partial: `director_credits`
    - [x] partial: `tv_credits`
    - [x] partial: `producer_credits`
    - [x] partial: `editor_credits`
    - [x] partial: `writer_credits`
    - [x] partial: `screenplay_credits`
    - [x] partial: `movie_credits`
- [x] `actor_more`
  - [x] erb: actor_more
    - [x] partial: `movie_credits`
    - [x] partial: `director_credits`
    - [x] partial: `tv_credits`
    - [x] partial: `producer_credits`
    - [x] partial: `editor_credits`
    - [x] partial: `writer_credits`
    - [x] partial: `screenplay_credits`

### DIFF
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -148,10 +148,6 @@ class TmdbController < ApplicationController
       @director_id = params[:director_id]
       @name = params[:name]
       @director = tmdb_handler_person_detail_search(@director_id)
-
-      @person_profile = @director.profile
-      @person_movie_credits = @director.movie_credits
-      @person_tv_credits = @director.tv_credits
     end
   end
 
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -82,7 +82,7 @@ class TmdbController < ApplicationController
   end
 
   def actor_more
-    @actor = tmdb_handler_actor_more(params[:actor_id])
+    @actor = tmdb_handler_person_detail_search(params[:actor_id])
   end
 
   def actor_credit
@@ -145,9 +145,7 @@ class TmdbController < ApplicationController
 
   def director_search
     if params[:director_id]
-      @director_id = params[:director_id]
-      @name = params[:name]
-      @director = tmdb_handler_person_detail_search(@director_id)
+      @director = tmdb_handler_person_detail_search(params[:director_id])
     end
   end
 
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -4,7 +4,6 @@ class TmdbController < ApplicationController
   require 'open-uri'
   include TmdbHandler
   include SearchParamParser
-  include TmdbHandler
 
   def search
     if @movie_title = params[:movie_title] || params[:movie_title_header]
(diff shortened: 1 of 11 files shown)


app/controllers/tmdb_controller.rb
```diff
@@ -4,6 +4,7 @@ class TmdbController < ApplicationController
   require 'open-uri'
   include TmdbHandler
   include SearchParamParser
+  include TmdbHandler
```
lortza, line 7:
```suggestion
```

app/views/tmdb/_director_credits.html.erb
```diff
@@ -1,13 +1,13 @@
-<h3>Director Credits (<%= @person_movie_credits.directing.size + @person_tv_credits.directing.size %>)</h3>
-  <p>Movies (<%= @person_movie_credits.directing.size %>) </p>
+<h3>Director Credits (<%= movie_credits.directing.size + tv_credits.directing.size %>)</h3>
```
lortza, line 1: This file is an example of the trend you'll see in all of the partials: replace the instance variables with partial locals

app/controllers/tmdb_controller.rb
```diff
@@ -81,8 +81,7 @@ def two_actor_search
   end
 
   def actor_more
-    @actor_id = params[:actor_id]
-    tmdb_handler_actor_more(@actor_id)
+    @actor = tmdb_handler_person_detail_search(params[:actor_id])
```
lortza, line 84: The `tmdb_handler_actor_more` id just a wrapper around `tmdb_handler_person_detail_search`. Nothing else calls it. I got rid of it.

lib/tmdb_handler.rb
```diff
@@ -128,22 +128,22 @@ def self.tmdb_handler_update_movie(movie)
...
     api_tv_credits_url = "#{BASE_URL}/person/#{person_id}/tv_credits?api_key=#{ENV['tmdb_api_key']}"
     tv_credits_results = JSON.parse(open(api_tv_credits_url).read, symbolize_names: true)
-    @person_tv_credits = TVPersonCredits.parse_result(tv_credits_results)
+
+    OpenStruct.new(
+      person_id: person_id,
+      profile: MoviePersonProfile.parse_result(bio_results),
+      movie_credits: MoviePersonCredits.parse_result(movie_credits_results),
+      tv_credits: TVPersonCredits.parse_result(tv_credits_results)
+    )
```
lortza, line 146: It may be more Railsy to create another proper class for this object, but I really don't think we need it because we're not doing anything else with this data. It's just a little package being sent directly to a display controller and isn't doing anything more data-intense.

mikevallano (approved):
👏 🚀 🎉 🍾

PR #267, opened 2022-01-01
Reduce instance variables in search and two_movie_search actions

lortza:
## Problems Solved
This PR is chipping away at the instance variables in the `TmdbHandler` and related controller. The focus of this work is on the movie `search` and `two_movie_search`.  Most of the files changed are doing the same change to where the `movie_partial_loop` partial is rendered.

It also resolves the bug of this button showing up twice:
Screen Shot 2022-01-01 at 4 32 39 PM

## Risk
Similar to #265, I've stepped through all of the workflows in the browser and kept a detailed list of the code I needed to trace to ensure everything still worked. Between that and specs, I'm hoping we're good to go here.

## TODO:
- [x] `tmdb_controller#search`
- [x] `Tmdb#tmdb_handler_search`
  - [x] check where `tmdb_handler_search` is called
  - [x] instance variables created here
    - `@query`
    - `@search_url`
    - `@tmdb_response`
    - `@discover_results`
    - `@not_found`
    - `@movies`
  - [x] erb: `tmdb/search`
  - [x] erb: `movie_partial_loop`
 - [x] where `movie_partial_loop` appears
   - [x] erb: `lists/public_show`
   - [x] erb: `lists/show`
   - [x] erb: `movies/index`
   - [x] erb: `tmdb/actor_search`
   - [x] erb: `tmdb/discover_search`
   - [x] erb: similar movies <-- going away
   - [x] erb: `tmdb/two_actor_search`
- [x] `two_movie_search` (calls `tmdb_handler_search`)
  - [x] reconcile instance variables dumped in here from `tmdb_handler_movie_more`


app/controllers/tmdb_controller.rb
```diff
@@ -7,8 +7,13 @@ class TmdbController < ApplicationController
 
   def search
     if @movie_title = params[:movie_title] || params[:movie_title_header]
-      @movie_title = I18n.transliterate(@movie_title)
-      tmdb_handler_search(@movie_title)
+      @search_results = tmdb_handler_search(@movie_title)
```
lortza, line 10: Tucks all instance variables needed by the view into this single object and single variable.

app/controllers/tmdb_controller.rb
```diff
@@ -7,8 +7,13 @@ class TmdbController < ApplicationController
...
+    end
+  end
+
+  def two_movie_search
```
lortza, line 14: I moved this method up here because they're related.

app/views/lists/public_show.html.erb
```diff
@@ -2,7 +2,7 @@
 <br>
 
   <div id ="public_search_loop">
-    <%= render partial: 'movies/movie_partial_loop', object: @movies %>
+    <%= render partial: 'movies/movie_partial_loop', locals: { movies: @movies } %>
```
lortza, line 5: To ensure expected behavior, I switched the way we call this partial in every file to ensure we're using local variables.

app/views/movies/_movie_partial_loop.html.erb
```diff
@@ -1,28 +1,27 @@
+<div class="row tile-container">
```
lortza, line 1: This looks like a lot of changes, but it's just blankspaces. The main change is switching `@movies` to `movies` on line 2.

app/controllers/tmdb_controller.rb
mikevallano: 🎉 ❤️

app/views/tmdb/search.html.erb
```diff
@@ -1,23 +1,22 @@
...
+  <br>
 
 <% else %>
 
-  <% if @not_found.present? %>
-      <h2>Hmm. <%= @not_found %></h2>
-      <p>Check your spelling and try again.</p>
-      <p class="button-main"><%= link_to "Search Again", api_search_path %> </p>
+  <% if @search_results.not_found_message.present? %>
+    <h2>Hmm. <%= @search_results.not_found_message %></h2>
```
mikevallano, line 14: 😍

mikevallano (approved):
🚀 🏆

Issue #270, opened 2022-01-02
Reduce instance variables in TmdbHandler and tmdb_controller


lortza:
While implementing some new features recently that we have a lot of instance variables being generated by the `lib/tmdb_handler.rb` file. This is problematic because they end up way down stream in places we do not expect. This is a big job and we're chipping away at it a few methods at a time.

### `TmdbHandler` Methods to refactor 
- [x] `tmdb_handler_search` PR #267
- [x] `tmdb_handler_movie_autocomplete` PR #269
- [x] `tmdb_handler_person_autocomplete` PR #269
- [x] [`tmdb_handler_movie_more`](https://github.com/mikevallano/tmdb-moviequeue/issues/268) PR #277
- [x] [`tmdb_handler_full_cast`](https://github.com/mikevallano/tmdb-moviequeue/issues/271) PR #272
- [x] [`tmdb_handler_add_movie`](https://github.com/mikevallano/tmdb-moviequeue/issues/273) PR #276
- [x] `self.tmdb_handler_update_movie`
- [x] `tmdb_handler_person_detail_search`
- [x] `tmdb_handler_actor_credit` PR #252
- [x] `tmdb_handler_tv_series_search`
- [x] `tmdb_handler_tv_series_autocomplete`
- [x] `tmdb_handler_tv_series` PR #251
- [x] `tmdb_handler_tv_season` PR #251
- [x] `tmdb_handler_tv_episode` PR #254
- [x] `tmdb_handler_two_movie_search` PR #267
- [x] [`tmdb_handler_discover_search`](https://github.com/mikevallano/tmdb-moviequeue/issues/274)
  - PR #279
  - PR #298
  - PR #291
  - PR #296

PR #272, opened 2022-01-02
Reduce instance variables in tmdb_handler_full_cast

### LINKED ISSUE #271: Reduce instance variables in tmdb_handler_full_cast
Relates to #270

This is an outline of the scope of work needed to reduce the instance variables generated by and perpetuated by the `tmdb_handler_full_cast` method:

- [ ] tmdb_handler_full_cast
  - generates these instance variables:
    * `@movie_url`
    * `@result`
    * `@credits`
    * `@cast`
    * `@actors`
    * `@crew`
    * `@director_credits`
    * `@directors`
    * `@editor_credits`
    * `@editors`
  - inherits instance variables from
    - [ ] tmdb_handler_movie_more
    - [ ] MovieCast.parse_results
    - [ ] MovieDirecting.parse_results
    - [ ] MovieEditing.parse_results
  - is called by:
    - [ ] tmdb_controller#full_cast
      * renders in views:
       - [ ] full_cast.html.erb
        * `@movie`
        * `@actors`
        * `@directors`
        * `@editors`

lortza:
## Related Issues & PRs
Closes #271
Relates to #270

## Problems Solved
This PR cleans up the instance variables in the `tmdb_handler_full_cast` method. It is imperfect, as there is related work to clean up the `tmdb_handler_movie_more` method. However, this does chip away a little bit more at our instance variable reduction goal outlined in #270.

## Risk
This method is only called in once place, which is the tmdb_controller. Data from this method is only rendered in a single view (full_cast.html.erb) and there are no partials in that view. It is safe to remove any of the instance variables that are not used in these views or subsequent partials. In addition, even though this method inherits a ton of instance variables from these methods

```
tmdb_handler_movie_more
MovieCast.parse_results
MovieDirecting.parse_results
MovieEditing.parse_results
```

The only 4 that are used in the view are `@movie`, `@actors`, `@directors`, and `@editors`. While it is untidy to let the other ones float around in the universe, they don't post an immediate risk and can be cleaned up in subsequent PRs.

## Browser testing
There is only one view that this change affects. Phew! So you can verify easily by going to any movie's "Full Cast" page. 
- [x] http://localhost:3000/tmdb/full_cast?tmdb_id=23483

### DIFF
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -47,10 +47,9 @@ class TmdbController < ApplicationController
 
   def full_cast
     if params[:tmdb_id]
-      @tmdb_id = params[:tmdb_id]
-      tmdb_handler_full_cast(@tmdb_id)
+      @cast = tmdb_handler_full_cast(params[:tmdb_id])
     end
-  end #full cast
+  end
 
   def actor_search
     if params[:actor]
--- a/app/views/tmdb/full_cast.html.erb
+++ b/app/views/tmdb/full_cast.html.erb
@@ -6,43 +6,41 @@
   <div class="row">
     <div class="col-xs-8">
       <h2>Actors:</h2>
-
       <% @cast.actors.each do |actor| %>
-      <div class="headshot-container">
-      <%= link_to headshot_for(actor), actor_more_path(actor_id: actor.actor_id) %>
-        <div class="name-block">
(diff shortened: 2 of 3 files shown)


mikevallano (approved):
💥

lortza:
This PR feels a little bit like when the laundry pipe burst and we severed and capped the pipe so we could turn the water back on again. There is still a bunch of severed garbage out there, but right now, we can function cleanly without dealing with it directly.

Issue #274, opened 2022-01-02
Reduce instance variables in tmdb_handler_discover_search


lortza:
Relates to #270

This is an outline of the scope of work needed to reduce the instance variables generated by and perpetuated by the tmdb_handler_discover_search method:

- [ ] tmdb_handler_discover_search
  - generates these instance variables:
    * `@actor`
    * `@actor1_id`
    * `@actor1_url`
    * `@actor1_search_result`
    * `@actor2`
    * `@actor2_id`
    * `@actor2_url`
    * `@actor2_search_result`
    * `@year_search`
    * `@exact_year`
    * `@after_year`
    * `@before_year`
    * `@genre`
    * `@company`
    * `@mpaa_rating`
    * `@sort_by`
    * `@page`
    * `@year_select`
    * `@not_found`
    * `@people`
    * `@discover_url`
    * `@discover_results`
    * `@movies`
    * `@total_pages`
    * `@page`
    * `@previous_page`
    * `@next_page`
  - inherits instance variables from
    - [ ] MovieSearch.parse_results
  - is called by:
    - [ ] tmdb_controller#actor_search
      * renders in views:
        - [ ] actor_search.html.erb: `@actor`, `@actor2`, `@not_found`, `@movies`, `@page`, `@total_pages`, `@previous_page`, `@next_page`
          - [x] partial: movie_partial_loop: `@movies` as local
    - [ ] tmdb_controller#two_actor_search
      * renders in views:
        - [ ] two_actor_search.html.erb: `@actor`, `@not_found`, `@actor1_id`, `@movies`, `@page`, `@total_pages`, `@previous_page`, `@next_page`
          - [x] partial: movie_partial_loop: `@movies` as local
    - [ ] tmdb_controller#discover_search
      * renders in views:
        - [ ] discover_search.html.erb: `@movies`, `@not_found`, `@params_for_view`, `@page`, `@total_pages`, `@previous_page`, `@next_page`, `@sort_by`, `@date`, `@genre`, `@actor`, `@actor2`, `@company`, `@mpaa_rating`, `@year_select` 
          - [x] partial: movie_partial_loop: `@movies` as local

lortza:
Handled by #298

PR #279, opened 2022-01-08
Refactor tmdb_controller discover_search and params parsing

lortza:
## Related Issues & PRs
Relates to #270
Relates to #274

## Problems Solved
While refactoring the `tmdb_handler_discover_search` method, I disocvered that the controller action for `discover_search` that calls this method is quite complex. In order to better understand the instance variables being used, I chose to tackle the controller situation first. It ended up being a significant refactor of a few files. In order to reduce risk for this PR, I chose to stop this work here. I'll get back to the `tmdb_handler_discover_search` method refactor in a subsequent PR.

I recommend looking at the diff in "split" mode.


app/controllers/tmdb_controller.rb
```diff
@@ -15,7 +14,7 @@ def search
 
   def two_movie_search
     if params[:movie_one] && params[:movie_two]
-      @search_results = tmdb_handler_two_movie_search(params[:movie_one], params[:movie_two])
+      @search_results = tmdb_handler_search_common_actors_in_two_movies(params[:movie_one], params[:movie_two])
```
lortza, line 17: This rename isn't relevant to the rest of the work in this PR. I changed it because it was confusing us the other day and renaming it would make that pain go away.

lib/tmdb/client.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Tmdb
```
lortza, line 3: Reminder to myself: remove this from this PR and put it on another branch.

app/controllers/tmdb_controller.rb
```diff
@@ -134,53 +137,14 @@ def director_search
   end
 
   def discover_search
-    #format date/year hash passed in params. otherwise year is passed directly on paginated pages
-    params[:year] = params[:date][:year] if params[:date].present?
+    form_params = %i[sort_by date genre actor actor2 company mpaa_rating year_select page]
```
lortza, line 140: I considered putting these params (and the ones on line 146) in a private method, but decided against it. This controller has a lot of methods, so I wanted to encapsulate everything related to this method in this single space.

mikevallano (changes requested):
Awesome stuff! As discussed, we decided to break up the param parser between parsing for the API and parsing for the view.

lortza:
I just did a ton of testing hitting the api directly and in the browser. (FYI docs for this discover search https://developers.themoviedb.org/3/discover/movie-discover) and it's working as expected. I'm feeling gaslit because what in the heck was i experiencing before?? This makes me think that either the api is flaky or we have a race condition or other conflict with our instance variables. I'll be refactoring that api-interfacing method in my next PR, so any race condition we have will be removed.

lib/search_param_parser.rb
```diff
@@ -1,34 +1,71 @@
...
+            .join(', ')
     end
 
-  end #parase params
+    private
 
-end #final
+    def use_exact_year?(year_select)
+      year_select.blank? || year_select == 'exact'
+    end
```
mikevallano, line 69: 🤔 This doesn't seem to be used anywhere

lib/search_param_parser.rb
```diff
@@ -1,34 +1,71 @@
...
-          @passed_params[:after_year] = "#{@year}-12-31"
-        elsif @year_select == "before"
-          @passed_params[:before_year] = "#{@year}-01-01"
+      year = params[:date][:year] if params[:date].present?
+      if year.present?
+        output[:year_display] = case params[:year_select]
+                                when 'before' then "before #{year}"
+                                when 'after' then "after #{year}"
+                                else
+                                  "from #{year}"
```
mikevallano, line 50:
```suggestion
        output[:year_display] =
          case params[:year_select]
          when 'before' then "before #{year}"
          when 'after' then "after #{year}"
          else
            "from #{year}"
          end
```

lib/search_param_parser.rb
```diff
@@ -1,34 +1,71 @@
...
+        sort_options = Movie::SORT_BY.to_h
+        sort_key = sort_options.key(params[:sort_by])
+        output[:sort_display] = "sorted by #{sort_key}"
+      end
 
-    if !@passed_params.keys.include?('sort_by')
-      @passed_params[:sort_by] = "revenue"
+      output.select { |_k, v| v.present? }
+            .values
+            .join(', ')
```
mikevallano, line 62:
```suggestion
      output
        .select { |_k, v| v.present? }
        .values
        .join(', ')
```

lib/search_param_parser.rb
lortza: whoops! you're right. it was just part of the process and i forgot to clean it up.

app/controllers/tmdb_controller.rb
```diff
@@ -134,53 +137,12 @@ def director_search
...
-    if @keys.include?("sort_by")
-      @sort_selected = show_params[:sort_by]
-      @sort_options = Movie::SORT_BY.to_h
-      @sort_key = @sort_options.key(@sort_selected)
-      @sort_display = "sorted by #{@sort_key}" if @keys.include?("sort_by")
-    end
-    "#{@actor_display} #{@genre_display} #{@rating_display} #{@year_display} #{@sort_display}"
+    searchable_params = SearchParamParser.parse_movie_params(passed_params)
+    tmdb_handler_discover_search(searchable_params)
+    @params_for_view = SearchParamParser.parse_movie_params_for_display(passed_params)
```
mikevallano, line 146: How do you feel about some other variable name?

Since the method returns a string, it might be clearer to name it something like `@parsed_results_text` so that it's not confused for more params?

app/controllers/tmdb_controller.rb
```diff
@@ -134,53 +137,12 @@ def director_search
...
-
-    if @keys.include?("sort_by")
-      @sort_selected = show_params[:sort_by]
-      @sort_options = Movie::SORT_BY.to_h
-      @sort_key = @sort_options.key(@sort_selected)
-      @sort_display = "sorted by #{@sort_key}" if @keys.include?("sort_by")
-    end
-    "#{@actor_display} #{@genre_display} #{@rating_display} #{@year_display} #{@sort_display}"
+    searchable_params = SearchParamParser.parse_movie_params(passed_params)
+    tmdb_handler_discover_search(searchable_params)
```
mikevallano, line 145: Since `searchable_params` is only used on the following line, how do you feel about skipping the variable and doing it like:

```ruby
    tmdb_handler_discover_search(
      SearchParamParser.parse_movie_params(passed_params)
    )
```

app/controllers/tmdb_controller.rb
lortza: `@params_for_view` is what is currently in-use in the view. I was trying to not touch the instance variables in the view for this PR. I'll totes change it in the next one where I mess around in the view though.

app/controllers/tmdb_controller.rb
lortza: it's not my fave. It feels like one think further than I want to go at a time, but if you feel strongly about it, we can change it.

mikevallano: 👍 fine w/ me

app/controllers/tmdb_controller.rb
mikevallano: 👍 Good plan!

mikevallano (approved):

Issue #273, opened 2022-01-02
Reduce instance variables in tmdb_handler_add_movie


lortza:
Relates to #270

This is an outline of the scope of work needed to reduce the instance variables generated by and perpetuated by the tmdb_handler_add_movie method:

- [x] tmdb_handler_add_movie
  - generates these instance variables:
    - [x] none! weee!
  - inherits instance variables from
    - [x] nowhere! weeee!
  - is called by:
    - [x] listings_controller#create
      * renders in views:
      - [x] create.js.erb: `@movie`, `@from`, `@listing`
        - [x] movie/movie_modal_innard
          - [x] movie_tag
          - [x] movie_on_lists
          - [x] movie_add_to_another_list
          - [x] movie_list_priority
          - [x] movie_add_to_first_list
          - [x] movie_rating
          - [x] movie_review
        - [x] movie/movie_show_list_manage: `@movie`
          - [x] movie_on_lists
          - [x] movie_add_to_another_list
          - [x] movie_add_to_first_list
        - [x] movie_on_lists
        - [x] movie_add_to_another_list
        - [x] movie_show_list_manage
    - [x] movies_controller#set_movie before: `update`
    - [x] screenings_controller#set_movie before each
      * renders in views:
        - [x] index.html.erb: `@movie`, `@screenings`
        - [x] index.json.jbuilder: `@screenings`
        - [x] new.html.erb: `@movie`
        - [x] create.js.erb: `@screening`
          - [x] movie_screening
          - [x] movie_rating
          - [x] movie_review
          - [x] movie_show_screening
          - [x] movie_rating
          - [x] movie_review
        - [x] _form.html.erb: `@movie`, `@screening`
        - [x] edit.html.erb: `@movie`, `@screening`
    - [x] taggings_controller#set_movie before each: `create`, `destroy`
      * renders in views:
        - [x] create.js.erb: `@movie`, `@tag_names`, `@list`
          - [x] movie_tag: `@list`
        - [x] destroy.js.erb: `@error`, `list`
          - [x] movie_tag: `@list`

PR #280, opened 2022-01-10
Build new discover_search method for tmdb_controller discover_search action

lortza:
## Related Issues & PRs
Relates to #270 
Relates to #273

## Problems Solved
The `tmdb_handler_discover_search` method is called by 3 controller actions. I've chosen to rewrite this method in a new `Tmdb::Client` and use the new method only with the `tmdb_controller#discover_search` action. This allows me to isolate the instance variable changes to just this action and view. The other 2 actions are calling the old method because those views still rely on the instance variables in that method. I'll do that work in a subsequent PR.

## Screenshots
[screenshots here]

## To Do
- [ ] check pagination is passing the right stuff and being processed correctly on all three search pages
- [ ] check error handling
  - [ ] when first arriving on page
  - [ ] when no results
  - [ ] when form not filled out completely
  - [ ] when actor autocomplete is skipped 
- [ ] update feature tests
- [ ] add unit tests
- [ ] look for ways to reduce calls to tmdb
- [ ] look for ways to reduce params domain knowledge


app/controllers/tmdb_controller.rb
```diff
@@ -136,13 +140,13 @@ def director_search
...
+  # advanced search
   def discover_search
-    form_params = %i[sort_by date genre actor actor2 company mpaa_rating year_select page]
+    form_params = %i[sort_by date genre actor actor2 company mpaa_rating year year_select page]
     passed_params = params.slice(*form_params).select { |_k, v| v.present? }
     return if passed_params.blank?
 
     searchable_params = SearchParamParser.parse_movie_params(passed_params)
-    tmdb_handler_discover_search(searchable_params)
-    @params_for_view = SearchParamParser.parse_movie_params_for_display(passed_params)
```
lortza, line 146: I moved this into a couple of different places:
* the original search params are returned in the `@results.original_search` so it can be accessed in the view
* the logic to build that big string is moved from the `SearchParamParser` into the `MovieHelper` since it is display/view logic

lib/tmdb_handler.rb
```diff
@@ -198,87 +198,4 @@ def tmdb_handler_search_common_actors_in_two_movies(movie_one_title, movie_two_t
       )
     end
   end
-
-  def tmdb_handler_discover_search(params)
```
lortza, line 202: This has been replaced by the new `Tmdb::Client.movie_advanced_search(params)` method

spec/lib/search_param_parser_spec.rb
```diff
@@ -41,225 +41,80 @@
...
-          )
-          expect(result[:exact_year]).to be(nil)
-          expect(result[:before_year]).to be(nil)
-          expect(result[:after_year]).to be(nil)
-        end
-      end
-    end
-  end
-
-  describe 'self.parse_movie_params_for_display' do
```
lortza, line 130: All of this is moved to the `spec/helpers/movies_helper_spec.rb` file

lortza:
Superseded by #298

PR #284, opened 2022-01-12
Refactor: Move tmbd handler methods into client

lortza:
## Related Issues & PRs
Relates to #270

## Problems Solved
We've been wanting to move the `TmdbHandler` methods out of a module and into a client class. This is part of the refactoring out the unpredictable nature of instance variables + module inclusion in our main controller. This PR moves all but 1 method out of that module and into a client class. The remaining method requires more refactoring and I'm handling that in PR #280.

I considered making smaller PRs -- one for each method being moved over -- but I didn't want to pepper you with busywork and I wanted to reduce the risk of this transition work not being completed (leaving us with code in 2 places).


.rubocop.yml
```diff
@@ -16,7 +16,7 @@ Metrics/BlockLength:
   Exclude:
     - !ruby/regexp /.*_spec\.rb$/
 Metrics/LineLength:
-  Max: 99
+  Max: 120
```
lortza, line 19: A line length of 120 feels reasonable to me.

app/controllers/listings_controller.rb
```diff
@@ -7,7 +7,7 @@ class ListingsController < ApplicationController
 
   def create
     @from = params[:from]
-    @movie = Movie.find_by(tmdb_id: params[:tmdb_id]) || tmdb_handler_add_movie(params[:tmdb_id])
+    @movie = GuaranteedMovie.find(params[:tmdb_id])
```
lortza, line 10: The `tmdb_handler_add_movie` felt funny to me -- like it didn't belong in the `Tmdb::Client` space or the `Movie` modal space. Then when I saw it in the context of where it is being used, I decided that it could live in a `GuaranteedMovie` class that also handles the `||` logic.

lib/tmdb/client.rb
```diff
@@ -0,0 +1,220 @@
...
+      def tv_episode(series_id:, season_number:, episode_number:)
+        episode_data = get_parsed_tv_episode_data(
+          series_id: series_id,
+          season_number: season_number,
+          episode_number: episode_number
+        )
+        TVEpisode.parse_record(episode_data)
+      end
+
+      private
```
lortza, line 162: These private methods are the closest to the API bone -- they hold the knowledge of the url string and do the JSON parsing. These methods are reused in the public methods above. Moving this logic into private methods made it easier to test the public methods because I could mock them.

lib/tmdb_handler.rb
```diff
@@ -9,196 +9,6 @@ def initialize(message)
...
-      OpenStruct.new(
-        movie_one: movie_one,
-        movie_two: movie_two,
-        common_actors: movie_one.actors & movie_two.actors,
-        not_found_message: nil
-      )
-    end
-  end
-
   def tmdb_handler_discover_search(params)
```
lortza, line 12: This last method has more refactoring work to do. I started that work in PR #280.

mikevallano (approved):
🎉 Beautiful!

Only one method (a beast, but only one method) remaining in tmdb handler and then it gets the 🪓 .

I created [a separate issue](https://github.com/mikevallano/tmdb-moviequeue/issues/288) that we can pair on to rename stuff (whether related to this PR or not) to make our lives easier.

PR #289, opened 2022-02-11
Consolidate tmdb client url methods

lortza:
After refactoring methods out of the `TmdbHandler` and into this `Tmdb::Client` space, we decided to restructure the way we interact with all of the urls required to us the TMDB API. This PR takes the approach of moving all urls into 1 main url-building method. This allows us to compartmentalize the concerns of url knowledge and parsing knowledge into the appropriate methods. It also make this file much tidier and much easier to read.

@mikevallano and I paired on this to determine the approach we wanted to take.

TODO:
- [x] reconcile all methods with old ones to make sure `[:results]` (etc) is in the right place
- [x] specs -- and maybe use VCR
- [x] add comments on `update_movie` about where we want to move it, etc.
- [x] look into dig vs fetch


lib/tmdb/client.rb
```diff
@@ -104,116 +108,134 @@ def common_actors_between_movies(movie_one_title, movie_two_title)
...
 
       def person_detail_search(person_id)
-        bio_results = get_parsed_person_bio(person_id)
-        movie_credits_results = get_parsed_person_movie_credits(person_id)
-        tv_credits_results = get_parsed_person_tv_credits(person_id)
+        person_params = { person_id: person_id }
+        person_data = request(:person_data, person_params)
+
+        # person_url = url_for_person_data(person_id)
+        # person_data = get_data(person_url)
```
lortza, line 122: @mikevallano This whole PR is super WIP. 
That said, most of this `person_detail_search` method uses the `url_for_...` convention we were talking about a couple of days ago. Then I went rogue and tried a different approach on lines 118 & 119. It corresponds with `request` method on 183 (which is just a start of the idea as well). It might be a little _too_ fancy. We'll start to see some grossness when we encounter the "discover search" url builder -- which i guess could be its own method... 🤔 

I just wanted to throw this out there as a visual aid for our discussion.

lib/tmdb/client.rb
```diff
@@ -104,116 +108,134 @@ def common_actors_between_movies(movie_one_title, movie_two_title)
...
-        url = "#{BASE_URL}/credit/#{credit_id}?api_key=#{API_KEY}"
+      def request(endpoint, params)
+        endpoints = {
+          credits_data: "/credit/#{params[:credit_id]}?api_key=#{API_KEY}",
+          person_data: "/person/#{params[:person_id]}?api_key=#{API_KEY}",
+          person_movie_credits: "/person/#{params[:person_id]}/movie_credits?api_key=#{API_KEY}"
+        }
+        url = "#{BASE_URL}#{endpoints[endpoint]}"
         JSON.parse(open(url).read, symbolize_names: true)
       end
```
lortza, line 191: This is just the start of what would be a single method to replace all of the `url_for...` methods.

app/controllers/movies_controller.rb
```diff
@@ -46,12 +46,12 @@ def show
 
   def modal
     @list = List.find(params[:list_id]) if params[:list_id].present?
-    @movie = Movie.find_by(tmdb_id: params[:tmdb_id]) || Tmdb::Client.movie(params[:tmdb_id])
+    @movie = GuaranteedMovie.find(params[:tmdb_id])
```
mikevallano, line 49: I don't think we want to make this change, and this a great reason why we should rename stuff. 

`GuaranteedMovie.find` actually does `find or create` where as `Client.movie` either returns the persisted movie or an instance of `MovieMore`, which is not backed by a db table.

So in the case where the user searches and sees results (which may or may not be stored in the db), when clicking the modal, this would save the movie to the db rather than displaying data. 

That's not necessarily a bad change, but it's very difficult to know that behavior is changing just by looking at this diff.

lib/tmdb/client.rb
mikevallano: 👍 Yeah, I really like this approach.

lib/tmdb/client.rb
```diff
@@ -92,8 +97,8 @@ def common_actors_between_movies(movie_one_title, movie_two_title)
         if not_found_message.present?
           OpenStruct.new(not_found_message: not_found_message)
         else
-          movie_one = GuaranteedMovie.find_or_initialize_from_api(movie_one_results.movies.first.tmdb_id)
-          movie_two = GuaranteedMovie.find_or_initialize_from_api(movie_two_results.movies.first.tmdb_id)
+          movie_one = get_movie_data(movie_one_results.movies.first.tmdb_id)
+          movie_two = get_movie_data(movie_two_results.movies.first.tmdb_id)
```
lortza, line 101: We can talk about making these api calls more performant. Maybe there are more `append_to_resutls...` things we can add. I don't want the client to know about the database though unless we're talking about some pretty clear caching.

lib/tmdb/client.rb
```diff
@@ -104,116 +109,105 @@ def common_actors_between_movies(movie_one_title, movie_two_title)
       end
 
       def person_autocomplete(query)
-        data = get_parsed_multi_search_results(query)
-        person_results = data.select{ |result| result[:media_type] == "person"}
-        person_results.map{ |result| result[:name] }.uniq
+        data = request(:multi_search, query: query)&.dig(:results)
+        data.select { |result| result[:media_type] == 'person' }
+            .map { |result| result[:name] }
+            .uniq
```
lortza, line 115: I'd like to clean this up a bit more.

lib/tmdb/client.rb
lortza: This evening at Ruby Tuesday, @hayesmp  gave me the idea i think we were reaching for here  ❤️ 🙌  -- make this hash a case statement instead. This way we're only building the url we want instead of building all the urls when building the hash.

lib/tmdb/client.rb
```diff
@@ -53,38 +51,46 @@ def get_movies_for_actor(actor_name:, page:, sort_by:)
...
         )
       end
 
       def movie_autocomplete(query)
-        data = get_parsed_movie_search_results(query)
+        data = request(:movie_search, query: query)[:results]
         data.map { |d| d[:title] }.uniq
       end
 
+      # TODO: Move this to the Movie class as update_with_api_data
```
mikevallano, line 77: 👍 I created issue #294 to make sure we don't forget about it

lib/tmdb/client.rb
```diff
@@ -1,40 +1,38 @@
...
           movies: movies
         )
       end
 
       def get_movies_for_actor(actor_name:, page:, sort_by:)
-        person_url = url_for_person_search(actor_name)
-        person_data = request_data_from_api(person_url)&.dig(:results)&.first
+        person_data = request(:person_search, query: actor_name)[:results]&.first
 
         return OpenStruct.new(not_found_message: "No actors found for '#{actor_name}'.") if person_data.blank?
```
mikevallano, line 26: Nitpick. Can we break this up instead of a one-liner?

```ruby
        return OpenStruct.new(
          not_found_message: "No actors found for '#{actor_name}'."
        ) if person_data.blank?
```

mikevallano (approved):
🎉 😍 Looks great. Thanks!

PR #291, opened 2022-02-12
Break actor search out of discover search

lortza:
The `TmdbHandler#tmdb_handler_discover_search` houses a few different actions. This PR starts breaking apart some of those actions into their own concerns by moving the single actor search into its own method called `Tmdb::Client.get_movies_for_actor`. 

Other changes:
* implement a `searchable_query(query)` method to handle the case when a user enters non-alphanumeric characters in to a search field. 
* start implementing the `url_for...` and `request_data_from_api` approach to client api calls, though I think eventually, we'll choose another direction from #289 's explorations


lib/tmdb/client.rb
```diff
@@ -201,20 +233,50 @@ def get_parsed_tv_episode_data(series_id:, season_number:, episode_number:)
...
+      def url_for_person_search(person_name)
+        searchable_name = I18n.transliterate(person_name)
+        "#{BASE_URL}/search/person?api_key=#{API_KEY}&query=#{searchable_name}"
+      end
+
+      def get_parsed_movie_search_results(movie_title)
+        searchable_title = I18n.transliterate(movie_title)
+        url = "#{BASE_URL}/search/movie?api_key=#{API_KEY}&query=#{searchable_title}"
         JSON.parse(open(url).read, symbolize_names: true)&.dig(:results)
       end
```
lortza, line 257: When I switched to the `person_url` & `person_data` approach from the `get_parsed_person_search_results` approach, I needed to move this `I18n.transliterate` method into all of these other places. It's not super relevant to the rest of the work in this PR, but I wanted to call it out because it you may otherwise wonder what in the heck is going on here.

lib/tmdb/client.rb
```diff
@@ -201,20 +233,50 @@ def get_parsed_tv_episode_data(series_id:, season_number:, episode_number:)
...
+
+      def url_for_movie_discover_search(params)
```
lortza, line 259: This is the exciting new way to build a url for the discover search 😃

mikevallano: 😃

app/controllers/movies_controller.rb
```diff
@@ -46,12 +46,12 @@ def show
 
   def modal
     @list = List.find(params[:list_id]) if params[:list_id].present?
-    @movie = Movie.find_by(tmdb_id: params[:tmdb_id]) || Tmdb::Client.movie(params[:tmdb_id])
+    @movie = Movie.find_by(tmdb_id: params[:tmdb_id]) || Tmdb::Client.get_movie_data(params[:tmdb_id])
```
mikevallano, line 49: Doobelive these can use `GuaranteedMovie.find_or_initialize_from_api`, but the other PR is now merged, so merging/rebasing should give you answers.

app/controllers/tmdb_controller.rb
```diff
@@ -53,13 +53,12 @@ def full_cast
...
-      params[:sort_by] = params[:sort_by] || "popularity"
-
-      # TODO: move to separate query
-      tmdb_handler_discover_search(params)
+    if params[:actor].present?
+      @results = Tmdb::Client.get_movies_for_actor(
+        actor_name: params[:actor],
+        page: (params[:page].presence || 1),
+        sort_by: (params[:sort_by].presence || 'popularity')
+      )
```
mikevallano, line 61: 😃 So much better!

lib/tmdb/client.rb
```diff
@@ -21,21 +21,53 @@ def movie_search(movie_title)
...
 
-      def movie(movie_id)
-        data = get_parsed_movie_data(movie_id)
-        MovieMore.parse_result(data)
+      def get_movies_for_actor(actor_name:, page:, sort_by:)
+        person_url = url_for_person_search(actor_name)
+        person_data = get_data(person_url)&.dig(:results)&.first
+        if person_data.blank?
+          OpenStruct.new(not_found_message: "No person matches found for '#{actor_name}'.")
+        else
```
mikevallano, line 29: How do you feel about returning the OpenStruct here rather than the if/else?

lib/tmdb/client.rb
```diff
@@ -21,21 +21,53 @@ def movie_search(movie_title)
...
+          movie_url = url_for_movie_discover_search(
+            people: person_data[:id],
+            page: page,
+            sort_by: sort_by
+          )
+          movie_data = get_data(movie_url)
+          movie_results = movie_data&.dig(:results)
```
mikevallano, line 36: Do we need the separate `movie_data` variable? It doesn't look like it's used anywhere else.

lib/tmdb/client.rb
```diff
@@ -201,20 +233,50 @@ def get_parsed_tv_episode_data(series_id:, season_number:, episode_number:)
...
+          url += "&primary_release_year=#{params[:year]}" if params[:year_select] == 'exact'
+          url += "&primary_release_date.lte=#{params[:year]}-01-01" if params[:year_select] == 'before'
+          url += "&primary_release_date.gte=#{params[:year]}-12-31" if params[:year_select] == 'after'
+        elsif params[:year].present?
+          url += "&primary_release_year=#{params[:year]}"
+        end
+        url
+      end
+
+      def get_data(url)
```
mikevallano, line 277: 🤔 I'm wondering about a clearer name here to indicate it's doing the actual api request.

lib/tmdb/client.rb
lortza: It's used on line 37. It's annoying, but we need the pagination info and it's at the same level as the `:results` key:
```
movie_data = get_data(movie_url)
movie_results = movie_data&.dig(:results) <--
total_pages = movie_data&.dig(:total_pages) <--
```

lib/tmdb/client.rb
lortza: I can see what you mean. 🤔 I think we're exploring that totally different approach like `request(:person_search, params)` in PR #289. I'm hoping to chip away at this `tmdb_handler_discover_search` while making minimal changes to the other methods, then we can go back and replace all the api calls with that `request` method.

lortza:
Now that all specs are passing, @mikevallano, I formally re-request your re-review. 😄

lib/tmdb/client.rb
lortza: Done.

mikevallano (approved):
🎉

Issue #288, opened 2022-02-09
Refactor naming to make things clearer


mikevallano:
The way the setup is with having persisted movies vs tmdb results, and all of the code surrounding it could use some refactoring.

lortza:
Are you starting this branch or am I?

lortza:
We handled a lot of the `Tmdb::Client` clarification in PR #295. I believe we still had some other items to rename inside the parsing classes like `MovieMore`.

mikevallano:
Leaving a note here from our email to consider Null Object patter when dealing with movies that may or may not be in the DB

PR #295, opened 2022-02-21
Rename Tmdb::Client methods to be more idiomatic

lortza:
## Problems Solved
Related to Issue #288

Our goal with this work is to make the methods in our main API client more idiomatic for API-facing names. For example,  `movie_autocomplete` is the name of a feature in our application, but it is not what that client method does. The client method returns a list of movie titles based on a query. Calling our client method `get_movie_titles` is more appropriate because it is `get`ting to the API and returning a list of movie titles. 

I paired with @mikevallano on the concepts we wanted to use for these names.

We're hoping this rename will clarify the purpose of these methods and make it easier to understand the code more quickly at a glance.

We're implementing these kinds of changes in a series of small PRs to reduce risk and to make each PR easier to review.

### DIFF
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -88,7 +88,7 @@ class TmdbController < ApplicationController
     query = show_title = params[:show_title] || params[:show_title_header]
     if query.present?
       @query = I18n.transliterate(query)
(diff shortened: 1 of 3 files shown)


mikevallano (approved):
😍

PR #296, opened 2022-02-23
Break two actor search out of discover search

lortza:
## Related Issues & PRs
Relates to #270 
Relates to #280

## Problems Solved
The `TmdbHnadler#discover_search` has a lot going on. I'm in the process of refactoring it. In order to do that, I'm separating its concerns out into specialized, testable methods. This PR breaks the "two actor search" aka "common movies between 2 actors" search into its own method. In the process, I made it flexible enough to handle any number of actors, which is neat, and not a feature we have right now.


app/controllers/tmdb_controller.rb
```diff
@@ -51,27 +52,29 @@ def full_cast
...
-  # single actor search
+  # movies for single actor
   def actor_search
     if params[:actor].present?
       @results = Tmdb::Client.get_movies_for_actor(
         actor_name: params[:actor],
-        page: (params[:page].presence || 1),
-        sort_by: (params[:sort_by].presence || 'popularity')
+        page: params[:page],
+        sort_by: params[:sort_by]
```
lortza, line 61: I moved this logic into the client. I figured the query should know how to handle itself when it comes to defaults.

app/controllers/tmdb_controller.rb
```diff
@@ -51,27 +52,29 @@ def full_cast
...
-      params[:page] = params[:page] || 1
-      params[:sort_by] = params[:sort_by] || "popularity"
-
-      # TODO: move to separate query
-      tmdb_handler_discover_search(params)
+    if (params[:actor] && params[:actor2]) || params[:paginate_actor_names].present?
+      @results = Tmdb::Client.get_common_movies_between_multiple_actors(
+        actor_names: [
+          params[:actor],
+          params[:actor2]
```
lortza, line 72: For funsies, I made this method scalable. We're not locked into comparing only 2 actors anymore -- even thought the form only has 2 actor fields currently.

app/views/tmdb/two_actor_search.html.erb
```diff
@@ -1,46 +1,33 @@
 <% content_for(:title, "Search by Actors") %>
 
-<h1>Find Common Movies Between 2 Actors</h1>
-
-
-<% if !@actor.present? && !@actor2.present? && !@not_found.present? %>
-
-<p>Enter two actors and see the movies they've been in together.</p>
-
+<h1>Find Common Movies Between Actors</h1>
```
lortza, line 3: The main changes to this erb are around reducing the instance variable to a single `@results` object that has attributes. Though we're a few months out from our original mission of reducing instance variables in controllers/views/etc, here's where the rubber meets the road. :)

lib/tmdb/client.rb
```diff
@@ -21,11 +21,15 @@ def get_movie_search_results(movie_title)
...
         person_data = request(:person_search, query: actor_name)[:results]&.first
 
-        return OpenStruct.new(
-          not_found_message: "No actors found for '#{actor_name}'."
-        ) if person_data.blank?
+        if person_data.blank?
+          return OpenStruct.new(
+            not_found_message: "No actors found for '#{actor_name}'."
+          )
+        end
```
lortza, line 32: I know you prefer a `return` block with a trailing `if`, but rubocop converted it to this style. If you prefer, I can change it back.

lib/tmdb/client.rb
```diff
@@ -142,6 +146,56 @@ def get_common_actors_between_movies(movie_one_title, movie_two_title)
...
+          data = request(:person_search, query: name)[:results]&.first
+          if data.blank?
+            not_found_messages << "No actor found for '#{name}'."
+          else
+            person_ids << data[:id]
+            actor_names << data[:name]
+          end
+        end
+
+        if not_found_messages.present?
```
lortza, line 169: At each step of the way, we return a "not found" message if some step fails.

spec/features/tmdb_feature_spec.rb
```diff
@@ -93,26 +95,26 @@
 
       scenario "two movies search indicates first movie not found if search is bad" do
         visit(two_movie_search_path)
-        @movie_query1 = "sdlfkjsdflkjsdf"
-        @movie_query2 = "The Big Lebowski"
```
lortza, line 97: While I was in here cleaning up, I fixed these instance variables too.

app/models/movie_search.rb
```diff
@@ -40,5 +40,3 @@ def self.parse_results(json)
   end
 
 end
-
```
lortza, line 43: I've added another todo item for us to revisit this `MovieSearch` class.

lib/tmdb/client.rb
mikevallano: I actually don't have a strong preference, so 👍

lib/tmdb/client.rb
```diff
@@ -142,6 +146,56 @@ def get_common_actors_between_movies(movie_one_title, movie_two_title)
...
+        page = page.presence || 1
+        sort_by = sort_by.presence || 'popularity'
+        names = actor_names.uniq.reject { |name| name == '' }.compact.presence || paginate_actor_names.presence.split(';')
+        return if names.blank?
+
+        not_found_messages = []
+        person_ids = []
+        actor_names = []
+
+        actor_results ||= names.compact.each do |name|
```
mikevallano, line 159: 🤔 is `actor_results` being used anywhere?

app/views/tmdb/two_actor_search.html.erb
mikevallano: > rubber meets the road
🍑

app/controllers/tmdb_controller.rb
```diff
@@ -51,27 +52,29 @@ def full_cast
...
-    if params[:actor] && params[:actor2]
-      params[:actor] = I18n.transliterate(params[:actor])
-      params[:actor2] = I18n.transliterate(params[:actor2])
-      params[:page] = params[:page] || 1
-      params[:sort_by] = params[:sort_by] || "popularity"
-
-      # TODO: move to separate query
-      tmdb_handler_discover_search(params)
+    if (params[:actor] && params[:actor2]) || params[:paginate_actor_names].present?
+      @results = Tmdb::Client.get_common_movies_between_multiple_actors(
```
mikevallano, line 69: Depending on other similar uses, and perhaps not in this PR, but maybe we could come up with a better variable name than `results`.

mikevallano (commented):
Aside from a couple of minor things, it looks real good!

lib/tmdb/client.rb
lortza: Oh yeah! 🤦‍♀️  Great catch! This is a relic of a previous version before I switched to hoisting those arrays. I'll remove it.

app/controllers/tmdb_controller.rb
lortza: Yeah, this name kind of got lodged into my brain because we've been using that word _everywhere_. 😆 I'm deep in the weeds over here. 🌱🌱🌱🌱🌱  

There are other controller actions that use that pattern. Maybe we can do a separate PR where we rename those variables.

mikevallano (approved):
🎉

Issue #297, opened 2022-02-24
Make Tmdb::Client thinner


lortza:
I got to have a good discussion at work today about what belongs in a client and what doesn't. I think it would be good to bring this idea back to our project so we can explore this idea. In a nutshell, the client knows about the API. In our case, the `Tmdb::Client` knows how to interface with TMDB. We make a new service object that doesn't know anything about TMDB, but does know how to get data to the controller. If some day we swap out TMDB or pick up another API, this service wouldn't really care all that much and the controller doesn't care at all.

## Keep in the client
* the `request` method
* any private methods that support the request method

## Move to a Service object
* any of the `get_...` methods

## Change the controller
* to call the service object methods because it doesn't know or care whatsoever what API we use

PR #302, opened 2022-03-13
Move update_movie out of Tmdb::Client

lortza:
## Related Issues & PRs
* Relates to #294
* Relates to #297 

## Problems Solved
@mikevallano and I have been talking about how the `update_movie` didn't belong in the `Tmdb::Client` since it deals with managing an active record object. It also didn't really belong in the `Movie` model since it reaches beyond the scope of active record object itself (#294). It is a hybrid method that needs some kind of service.

We've also been talking about making our 🔥 new `Tmdb::Client` much skinnier (#297) -- housing only the `request` method and moving the `get_movie_titles...` type of methods into a data service of some sort.

This PR starts that process. 
* introduce a new `MovieDataService` object
* move `update_movie` into it
* move large arrays that are meant to help with form submissions (sorting options, genres, etc) from the `Movie` model into the `MovieDataService` object.

Future work I think will look like:
* moving other `get_movie...` methods from `Tmdb::Client` to `MovieDataService`
* moving TV-related methods from `Tmdb::Client` to `TvDataService`
* moving person-related methods from `Tmdb::Client` to `PersonDataService`

### DIFF
--- a/app/models/movie.rb
+++ b/app/models/movie.rb
@@ -16,35 +16,9 @@ class Movie < ActiveRecord::Base
   has_many :reviews
   has_many :ratings
   has_many :screenings
-  has_many :viewers, :through => :screenings,
-  :source => :user
+  has_many :viewers, :through => :screenings, :source => :user
 
   attr_accessor :production_companies
-
-  LIST_SORT_OPTIONS = [ ["title", "title"], ["shortest runtime", "shortest runtime"],
-  ["longest runtime", "longest runtime"], ["newest release", "newest release"],
-  ["vote average", "vote average"], ["recently added to list", "recently added to list"],
-  ["watched movies", "watched movies"], ["unwatched movies", "unwatched movies"],
-  ["recently watched", "recently watched"], ["highest priority", "highest priority"],
-  ["only show unwatched", "only show unwatched"], ["only show watched", "only show watched"] ]
-
-  MY_MOVIES_SORT_OPTIONS = [ ["title", "title"], ["shortest runtime", "shortest runtime"],
-  ["longest runtime", "longest runtime"], ["newest release", "newest release"],
-  ["vote average", "vote average"], ["watched movies", "watched movies"], ["recently watched", "recently watched"],
-  ["unwatched movies", "unwatched movies"], ["only show unwatched", "only show unwatched"],
-  ["only show watched", "only show watched"], ["movies not on a list", "movies not on a list"] ]
-
-  GENRES = [["Action", 28], ["Adventure", 12], ["Animation", 16], ["Comedy", 35], ["Crime", 80],
-  ["Documentary", 99], ["Drama", 18], ["Family", 10751], ["Fantasy", 14], ["Foreign", 10769], ["History", 36],
-  ["Horror", 27], ["Music", 10402], ["Mystery", 9648], ["Romance", 10749], ["Science Fiction", 878], ["TV Movie", 10770],
-  ["Thriller", 53], ["War", 10752], ["Western", 37]]
-
-  MPAA_RATINGS = [ ["R", "R"], ["NC-17", "NC-17"], ["PG-13", "PG-13"], ["G", "G"] ]
-
-  SORT_BY = [ ["Popularity", "popularity"], ["Release date", "release_date"], ["Revenue", "revenue"],
-  ["Vote average", "vote_average"], ["Vote count","vote_count"] ]
-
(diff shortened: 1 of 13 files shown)


lortza:
@mikevallano some questions:

1. I have not investigated the use of HTTParty in this method. Do you think that work belongs here or in a subsequent PR?
2. What do you think of this approach in general? Do you want to pop this thing into the `Movie` model instead?

app/models/movie.rb
```diff
@@ -16,36 +16,10 @@ class Movie < ActiveRecord::Base
   has_many :reviews
   has_many :ratings
   has_many :screenings
-  has_many :viewers, :through => :screenings,
-  :source => :user
+  has_many :viewers, :through => :screenings, :source => :user
 
   attr_accessor :production_companies
 
-  LIST_SORT_OPTIONS = [ ["title", "title"], ["shortest runtime", "shortest runtime"],
```
lortza, line 24: All of these arrays have been moved to the `MovieDataService`.

app/models/movie.rb
```diff
@@ -126,8 +101,8 @@ def priority(list)
   end
 
   def priority_text(list)
-    p = listings.find_by(list_id: list.id).priority
-    case p
+    priority = listings.find_by(list_id: list.id)&.priority
```
lortza, line 104: `p` feels reserved to me. If we ever wanted to do any debugging in this method and wanted to output the value of `p`, we may have a collision, so i renamed it to `priority`.

app/services/movie_data_service.rb
```diff
@@ -0,0 +1,74 @@
...
+  ["Thriller", 53], ["War", 10752], ["Western", 37]]
+
+  MPAA_RATINGS = [ ["R", "R"], ["NC-17", "NC-17"], ["PG-13", "PG-13"], ["G", "G"] ]
+
+  SORT_BY = [ ["Popularity", "popularity"], ["Release date", "release_date"], ["Revenue", "revenue"],
+  ["Vote average", "vote_average"], ["Vote count","vote_count"] ]
+
+  YEAR_SELECT = [ ["Exact Year", "exact"], ["After This Year", "after"], ["Before This Year", "before"] ]
+
+  def self.update_movie(movie)
```
lortza, line 29: I moved this method over and have done nothing else to it.

app/views/lists/show.html.erb
```diff
@@ -1,40 +1,36 @@
 <% content_for(:title, @list.name) %>
-<div class="row">
-
-<h1><%= @list.name %> <span class="count">(<%= pluralize(@list.movies.size, "Movie") %>) <icon class="<%= public_private_indicator(@list) %>"><icon></span></h1>
-<p>by <%= @list.owner.username %></p>
 
-<% if @list.description.present? %>
-  <p><%= @list.description %></p>
-<% end %> <!-- #if description.present -->
+<div class="row">
```
lortza, line 3: This is just blankspace cleanup & fixing the indention.

app/views/lists/show.html.erb
```diff
@@ -1,40 +1,36 @@
...
+  <h1><%= @list.name %> <span class="count">(<%= pluralize(@list.movies.size, "Movie") %>) <icon class="<%= public_private_indicator(@list) %>"><icon></span></h1>
+  <p>by <%= @list.owner.username %></p>
 
+  <% if @list.description.present? %>
+    <p><%= @list.description %></p>
+  <% end %> <!-- #if description.present -->
 
   <%= form_tag user_list_path(@list.owner, @list), { class: "form-class", id: "list-show-sort", role: "form", method: :get } do %>
-    <%= select_tag :sort_by, options_for_select(Movie::LIST_SORT_OPTIONS, @sort_by), { class: "form-control index-sort-dropdown", :include_blank => "Sort by", id: "list_sort_options" } %>
+    <%= select_tag :sort_by, options_for_select(MovieDataService::LIST_SORT_OPTIONS, @sort_by), { class: "form-control index-sort-dropdown", :include_blank => "Sort by", id: "list_sort_options" } %>
```
lortza, line 12: Aside from this line where I've updated the source of the `LIST_SORT_OPTIONS` constant, the rest of the changes in this file are just blankspace corrections.

app/models/movie.rb
mikevallano: 👍 Totes

mikevallano (approved):
😍 So much better!

PR #305, opened 2022-03-13
Make Tmdb::Client thinner

### LINKED ISSUE #297: Make Tmdb::Client thinner

lortza:
## Related Issues & PRs
Closes #297

## Problems Solved
The `Tmdb::Client` got a little chunky as we built it out to replace the `TmdbHandler`. This PR strips the client down to only the methods that make the call to the TMDB API and moves all of the parsing methods out into their own data services:
* `MovieDataService`
* `TvSeriesDataService`
* `PersonDataService`

Now the `Tmdb::Client` is skinny and clear in its purpose. 😸 

This is exciting as it may open up some ideas for us on how we want to handle some of the related parsing classes.


mikevallano (approved):
🎉

PR #309, opened 2022-03-15
Add seen movies section to actor more page

lortza:
## Problems Solved
Sometimes I don't know how I know someone's face. I know I've seen them in movies, but I don't remember which. This exploratory PR adds a "movies seen" section to an actor's bio page. It does not address TV actors.

## Screenshots
Sandy B has been in 60 movies. These are the movies I've seen her in. 
Screen Shot 2022-06-14 at 7 28 23 PM
Screen Shot 2022-06-14 at 7 28 32 PM

## Things Learned
* It can be hard to figure out a clever solution for this kind of UX problem.
* Sometimes it is good to just :yolo: and see if the users (us) like the implementation.

### DIFF
--- a/app/views/tmdb/_seen_movie_credits.html.erb
+++ b/app/views/tmdb/_seen_movie_credits.html.erb
@@ -1,5 +1,5 @@
 <h3>
-  Movie's I've Seen
+  Movies I've Seen
 </h3>
 <ul>
   <% movie_credits.actor.each do |credit| %>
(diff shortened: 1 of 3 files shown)


app/views/tmdb/_seen_movie_credits.html.erb
```diff
@@ -0,0 +1,15 @@
+<h3>
+  Movie's I've Seen
```
lortza, line 2:
```suggestion
  Movies I've Seen
```

PR #321, opened 2022-11-11
230: upgrade ruby to 2.7.6

mikevallano:
## Related Issues & PRs
Closes #230

## Problems Solved
* Upgrade ruby to be able to upgrade heroku stack

## Screenshots
ruby-upgrade-server-running

## Things Learned
* If your circle config isn't valid, circle lets you edit it directly and does the linting to show you that it's valid, then you can commit in the interface and rerun the job 😎


Gemfile
```diff
@@ -42,16 +43,14 @@ group :development, :test do
   gem 'capybara-email'
   gem 'guard-rspec', require: false
   gem 'terminal-notifier-guard'
-  gem 'spring-commands-rspec'
-  gem 'spring'
```
mikevallano, line 46: These were causing errors in trying to load the rails console. I believe they're removed in a later version of rails anyway.

Gemfile
```diff
@@ -42,16 +43,14 @@ group :development, :test do
...
 end
 
 group :test do
     gem 'shoulda-matchers'
     gem 'capybara'
     gem 'database_cleaner'
     gem 'launchy'
-    gem 'selenium-webdriver'
+    gem 'webdrivers'
```
mikevallano, line 53: Newer gem to handle this

.circleci/config.yml
```diff
@@ -2,10 +2,13 @@
 # See: https://circleci.com/docs/2.0/configuration-reference
 version: 2.1
 
+orbs:
+  browser-tools: circleci/browser-tools@1.4.0
```
mikevallano, line 6: This is a thing from circle that fixed the webdriver problem: https://circleci.com/developer/orbs/orb/circleci/browser-tools

Gemfile
lortza: 👍 I also believe that is true

lortza (approved):
That monkeypatch stuff is kind of weird, but since you explained it, I get it.

Everything else looks right on. 👍

PR #324, opened 2022-11-12
323: upgrade to rails 5.1

### LINKED ISSUE #323: Upgrade to rails 5.1
We have numerous deprecation warnings for 5.1 as well

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-0-to-rails-5-1

mikevallano:
## Related Issues & PRs
Closes #323

## Problems Solved
* Upgrade to Rails 5.1 as a way to keep things updated

## Screenshots
5-1-in-production

## Things Learned
* Rails has a `rails app:update` command. Run it and hold on to your butts


app/controllers/registrations_controller.rb
```diff
@@ -71,6 +71,7 @@ def update
           :update_needs_confirmation : :updated
         set_flash_message :notice, flash_key
       end
+      # bypass_sign_in(user)
```
mikevallano, line 74: Temp. Need to try this out for a Devise deprecation warning

config/environments/production.rb
```diff
@@ -73,27 +90,19 @@
...
-    enable_starttls_auto: true,
-    user_name: 'apikey',
-    password: ENV['SENDGRID_API_KEY']
-  }
-
-  # ActionMailer Config
-  config.action_mailer.default_url_options = { :host => 'http://www.flicksonlists.com/' }
-  config.action_mailer.delivery_method = :smtp
-  config.action_mailer.perform_deliveries = true
-  config.action_mailer.raise_delivery_errors = false
```
mikevallano, line 91: I kept this in the file, but it was moved because I manually added it back after `app:update` removed it

app/controllers/registrations_controller.rb
```diff
@@ -71,7 +71,7 @@ def update
           :update_needs_confirmation : :updated
         set_flash_message :notice, flash_key
       end
-      sign_in resource_name, resource, bypass: true
+      bypass_sign_in(resource)
```
mikevallano, line 74: This was because of devise deprecation warning

app/views/devise/confirmations/new.html.erb
```diff
@@ -1,11 +1,11 @@
 <h2>Resend confirmation instructions</h2>
 
 <%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }) do |f| %>
-  <%= devise_error_messages! %>
+  <%= render "devise/shared/error_messages", resource: resource %>
```
mikevallano, line 4: Resolve devise deprecation warning

config/initializers/json.rb
```diff
@@ -0,0 +1,9 @@
+# monkey patch for the json gem to fix ruby keyword arg deprecation
+# https://github.com/flori/json/issues/399
+module JSON
+  module_function
+
+  def parse(source, opts = {})
+    Parser.new(source, **opts).parse
+  end
+end
```
mikevallano, line 9: I added this to fix ruby deprecation warning. The code is in the gem. See https://github.com/flori/json/issues/399

app/controllers/registrations_controller.rb
```diff
@@ -1,6 +1,6 @@
 class RegistrationsController < Devise::RegistrationsController
-  prepend_before_filter :require_no_authentication, only: [:new, :create, :cancel]
-  prepend_before_filter :authenticate_scope!, only: [:edit, :update, :destroy]
+  prepend_before_action :require_no_authentication, only: [:new, :create, :cancel]
+  prepend_before_action :authenticate_scope!, only: [:edit, :update, :destroy]
```
lortza, line 3: I didn't know what the difference between a `before_action` and a `prepend_before_action` was, [so I looked it up](https://stackoverflow.com/a/49791090/5009528).

A `prepend_before_action` runs before a `before_action`. There are no `before_action`s in this controller, so it seems like these prepends could just be regular `before_action`s. 🤔 

Since we're dealing with Devise stuff, and I am not sure of the order of loading or any including that may be happening between controllers, I feel hesitant to make that change. Of course, if the Devise docs say this is the way, then this is the way. I'll do a deeper dive on that in a bit. I've got 80 more files to investigate...

mikevallano: Yeah, I looked and that's the default Devise code: https://github.com/heartcombo/devise/blob/main/app/controllers/devise/registrations_controller.rb

config/application.rb
```diff
@@ -17,6 +17,8 @@
 
 module TmdbMoviequeue
   class Application < Rails::Application
+     # Initialize configuration defaults for originally generated Rails version.
+     config.load_defaults 5.1
```
lortza, line 21: awwww yeah!

config/environments/test.rb
```diff
@@ -25,19 +27,19 @@
...
   config.action_controller.allow_forgery_protection = false
+  config.action_mailer.perform_caching = false
 
   # Tell Action Mailer not to deliver emails to the real world.
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
-  # Randomize the order test cases are executed.
-  config.active_support.test_order = :random
```
lortza, line 35: To clarify what we just talked about, this setting is being removed here because this is the new default behavior. There will be no change in behavior by removing this line.

config/initializers/assets.rb
```diff
@@ -3,9 +3,12 @@
 # Version of your assets, change this if you want to expire all your assets.
 Rails.application.config.assets.version = '1.0'
 
-# Add additional assets to the asset load path
+# Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
+# Add Yarn node_modules folder to the asset load path.
+Rails.application.config.assets.paths << Rails.root.join('node_modules')
```
lortza, line 9: lol. `node_modules`.

config/initializers/json.rb
lortza: ah. interesting.

config/initializers/new_framework_defaults_5_1.rb
```diff
@@ -0,0 +1,14 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 5.1 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
+#
+# Read the Guide for Upgrading Ruby on Rails for more info on each option.
+
+# Make `form_with` generate non-remote forms.
+Rails.application.config.action_view.form_with_generates_remote_forms = false
```
lortza, line 10: If you plan to keep this setting, we need to move it into a config file. This `config/initializers/new_framework_defaults_5_1.rb` is a temporary file for use to help us turn on/off new settings for a slow upgrade rollout. We should plan to delete this file soon.

lib/tmdb/client.rb
```diff
@@ -11,7 +11,7 @@ def request(endpoint, params)
         return { results: [] } if params[:query].present? && searchable_query(params[:query]).empty?
 
         api_path = build_url(endpoint, params)
-        response = open("#{BASE_URL}#{api_path}").read
+        response = URI.open("#{BASE_URL}#{api_path}").read
```
lortza, line 14: How in the heck did this even work before?

lortza (commented):
There are just a few comments I have left that I'd like to chat about with you. Otherwise, it looks really good!

config/initializers/new_framework_defaults_5_1.rb
mikevallano: Thanks for insights. I had no idea! After reading up on it, it looks like it only impacts `form_with` (which we're not currently using) and not `form_for`.

If we get to a point where we add a new form, surely we'll check to see what it's doing and can address it then?

So I think this can just be skipped.

The only other line in this file is commented out, so I think I'll just remove this whole file?

lib/tmdb/client.rb
mikevallano: 🤷 Based on the deprecation warning, I think it was still calling `open` on `URI`

config/initializers/new_framework_defaults_5_1.rb
lortza: Yep! Sounds good to me.

lortza (approved):

PR #337, opened 2022-12-30
WIP add Blazer gem for easy SQLing of prod data

### LINKED ISSUE #335: Implement Blazer for easy SQL access
https://github.com/ankane/blazer

lortza:
## Related Issues & PRs
Closes #335 

This PR is WIP. TODO:
- [ ] I have some questions about how we want to configure things
- [ ] I need to look into how we handle our secrets
- [ ] I need to try this as a non-admin user

## Screenshots
Screen Shot 2022-12-30 at 10 57 10 AM

## Things Learned
* When gem docs are not well-written, it leaves me feeling so very uneasy

## Due Dilligence Checks
- [ ] If this PR includes migrations, update seeds and factories
- [ ] I have written new tests for this work
- [ ] I have browser-tested this work
- [ ] If this PR involves access to secure information, I have put safety measures in place to protect it.
- [ ] I have successfully tested the feature branch in production.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem 'coveralls', '0.8.11', require: false
 gem 'rack-mini-profiler'
 gem 'stamp'
 gem 'sentry-raven'
+gem 'blazer' # SQL admin interface
 
 # Use ActiveStorage variant
 # gem 'mini_magick', '~> 4.8'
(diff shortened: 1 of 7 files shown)


mikevallano:
I tested it out with a non-admin and it properly restricts access (though weirdly `root_path` does too many redirects but using `'/'` works 🤷).

I also created a user locally with read-only permissions, as noted here: https://github.com/ankane/blazer#postgresql and it worked, though I want to look into that a bit more before doing it in production.

PR #338, opened 2022-12-30
Remove instance variables from api  data-parsing poros

lortza:
## Problems Solved
About a year ago I did a big push to refactor the `tmdb_controller` and related service objects with the goal of getting rid of rogue instance variables (as outlined in #270). I did _most_ of the work, but stopped just short of doing _all_ of the work. This PR cleans up the last of those API data-parsing POROs that had unnecessary instance variables in them.

This PR also cleans up a lot of spacing, unnecessary comments (like where a block ends), placement conventions, and changes `"` to `'` for easier rubocopping.

## Things Learned
* When we first built this app, naming was hard and staying organized was also hard. We're dealing with some of that weirdness today.
* All of the work we did last December really encapsulated those POROs, so when I went back for this refactor today, the tests were all green!

## Moving Forward
* I'd like to move our non-AR-backed models into another directory. [Bob Martin suggests](http://blog.vladimir-rybas.com/blog/2014/04/04/rails-and-pipes/)  using `entities` for nouns and `use_cases` for verbs.

### DIFF
--- a/app/models/movie_cast.rb
+++ b/app/models/movie_cast.rb
@@ -1,6 +1,9 @@
-class MovieCast
-  def initialize(actor_id, credit_id, name, character_name, profile_path, cast_id, order )
+# frozen_string_literal: true
 
+class MovieCast
+  attr_accessor :title, :actor_id, :credit_id, :name, :character_name, :profile_path, :cast_id, :order
+
+  def initialize(actor_id:, credit_id:, name:, character_name:, profile_path:, cast_id:, order:)
     @actor_id = actor_id
     @credit_id = credit_id
     @name = name
@@ -8,29 +11,19 @@ class MovieCast
     @profile_path = profile_path
     @cast_id = cast_id
     @order = order
+  end
 
-  end #init
-
-  attr_accessor :title, :actor_id, :credit_id, :name, :character_name, :profile_path, :cast_id, :order
-
-
-  def self.parse_results(json)
-    @actors = []
-    json.each do |result|
-      @actor_id = result[:id]
-      @credit_id = result[:credit_id]
-      @name = result[:name]
-      @character_name = result[:character]
-      @profile_path = result[:profile_path]
-      @order = result[:order]
-      @cast_id = result[:cast_id]
-
-      @actor = MovieCast.new(@actor_id, @credit_id, @name, @character_name, @profile_path, @order, @cast_id)
-      @actors << @actor
-    end #results loop
-    @actors
-  end #parse results
-
-end #class MovieCast
-
-
+  def self.parse_results(results)
+    results.map do |result|
+      new(
+        actor_id: result[:id],
+        credit_id: result[:credit_id],
+        name: result[:name],
+        character_name: result[:character],
+        profile_path: result[:profile_path],
+        order: result[:order],
+        cast_id: result[:cast_id]
+      )
+    end
+  end
+end
--- a/app/models/movie_credits.rb
+++ b/app/models/movie_credits.rb
@@ -1,7 +1,9 @@
+# frozen_string_literal: true
+
 class MovieCredits
+  attr_accessor :adult, :credit_id, :department, :job, :tmdb_id, :title, :poster_path, :date, :character
 
-  def initialize(adult, credit_id, department, job, tmdb_id, title, poster_path, date, character)
-
+  def initialize(adult:, credit_id:, department:, job:, tmdb_id:, title:, poster_path:, date:, character:)
     @adult = adult
     @credit_id = credit_id
     @department = department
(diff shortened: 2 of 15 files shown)


app/models/movie_credits.rb
```diff
@@ -11,32 +13,29 @@ def initialize(adult, credit_id, department, job, tmdb_id, title, poster_path, d
...
-      @credit = MovieCredits.new(@adult, @credit_id, @department, @job, @tmdb_id, @title, @poster_path, @date, @character)
-      @credits << @credit
-    end #loop
-    @credits = @credits.sort_by{ |credit| credit.date }.reverse
-  end #parse
+    private
 
-end #class
\ No newline at end of file
+    def format_release_date(date_data)
```
lortza: Since this method is private, I moved it into the `private` section to help us understand this object's interface.

app/models/movie_cast.rb
```diff
@@ -1,36 +1,29 @@
+# frozen_string_literal: true
```
lortza, line 1: This file is a good example of the changes I made to the other files in this PR. 
* move the `attr_accessor` to the top
* switch to keyword args
* use `map` to return a new array instead of shoveling into an array
* build the `new` object in the `map` instead of setting up all of the attributes first
* remove instance variables from the `map` block

app/models/movie_person_credits.rb
```diff
@@ -1,30 +1,25 @@
-class MoviePersonCredits
+# frozen_string_literal: true
 
-  def initialize(movies, tv, directing, editing, writing, screenplay, producer, actor)
```
lortza, line 3: Strangely enough, I don't think `movies` and `tv` were doing anything here. If you follow the code down the pipe, you'll see that the data for `movies` and `tv`is getting populated later. Perhaps the instance variables were created here in order to populate them later? Dunno. But they're gone now and stuff still works. 🧹

app/models/movie_search.rb
```diff
@@ -13,30 +14,27 @@ def initialize(title, in_db, tmdb_id, release_date, vote_average, overview, back
...
 
-          @movie = MovieSearch.new(@title, @in_db, @tmdb_id, @release_date, @vote_average, @overview, @backdrop_path, @poster_path, @tags, @lists)
-          @movies << @movie
-        end
-    end #loop
-    @movies
+      if existing_movie.present?
+        existing_movie
+      else
+        new(
```
lortza, line 25: It seems like there is probably a more elegant way to do this, but my brain got tired and this is how I left it. Any suggestions?

app/models/tv_credit_season.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
lortza, line 1: These are all cosmetic changes, but that is the nature of this PR.

spec/models/movie_person_profile_spec.rb
```diff
@@ -92,88 +92,89 @@
     end
   end
 
-  describe '.parse_bio' do
+  describe 'private.parse_bio' do
```
lortza, line 95: I chose to use `send` here in the specs instead of rewriting the public methods to account for these private methods. It's kind of a lazy move on my part. What's your preference for the codebase?

app/models/movie_search.rb
mikevallano: I mean maybe `Movie.find_by(tmdb_id: @tmdb_id) || new(` but I think the way you currently have it is more readable?

mikevallano (approved):
awwww yeah! much better code now. thanks!

app/models/movie_search.rb
```diff
@@ -13,30 +14,27 @@ def initialize(title, in_db, tmdb_id, release_date, vote_average, overview, back
...
+          title: result[:title],
+          in_db: false,
+          tmdb_id: tmdb_id,
+          release_date: result[:release_date],
+          vote_average: result[:vote_average]&.round(1),
+          overview: result[:overview],
+          backdrop_path: result[:backdrop_path],
+          poster_path: result[:poster_path],
+          tags: nil,
+          lists: nil
```
lortza, line 35: I'm wondering about these `tags` and `lists` and if we need these attributes at all. This object is not persisted and adding that data would require persistence... 🤔

app/models/movie_search.rb
lortza: I'm beginning to wonder why we're making a different object here and not just instantiating a `Movie` object. This `Movie` would probably not ever be saved. I know that later downstream in the `app/views/movies/_movie_partial_loop.html.erb` we ask the object it's class for the sake of routing:
```
<% if movie.is_a?(Movie) %>
  <%= link_to image_for(movie), movie_path(movie) %>
<% else %>
  <%= link_to image_for(movie), movie_more_path(tmdb_id: movie.tmdb_id) %>
<% end %>
```
so I'm thought-swirling a bit on how to do this whole concept in a better way. Not for this PR -- that would be a different refactor.

mikevallano: Yeah, sounds good for a different PR 👍

PR #341, opened 2023-01-04
Make MovieSearch object consistently return MovieSearch object 

lortza:
## Problem Solved
In the `MovieSearch` class, we were returning a list of movie-like objects, some of which were instances of `MovieSearch` and others were instances of `Movie` depending on whether the movie data had been persisted to the database. Because this list could container either class, we had a condition in the view that switched routes depending on the class type.

## Summary of Changes Made
In this PR, the `MovieSearch` (now called `MovieSearchResult`) class method returns only instances of `MovieSearchResult`.  As it turns out, we were already doing a persistence check in the controller via our `GuaranteedMovie` object, so we don't have to concern ourselves with doing it again. So now, instead of doing the persistence check up front in the `MovieSearch` parsing method and then asking about it later in the view, we now only do it in the controller.

I've renamed these object for clarity:
```ruby
# before
MovieSearch.parse_results(api_data)

#after
MovieSearchResult.build_list(api_data)
```

### DIFF
--- a/app/entities/movie_search_result.rb
+++ b/app/entities/movie_search_result.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class MovieSearch
+class MovieSearchResult
   attr_accessor :title, :in_db, :tmdb_id, :release_date, :vote_average, :overview, :backdrop_path, :poster_path, :tags, :lists
 
   def initialize(title:, in_db:, tmdb_id:, release_date:, vote_average:, overview:, backdrop_path:, poster_path:)
@@ -16,25 +16,18 @@ class MovieSearch
     @lists = nil
   end
 
-  def self.parse_results(results)
+  def self.build_list(results)
     results.map do |result|
-      tmdb_id = result[:id]
-      existing_movie = Movie.find_by(tmdb_id: @tmdb_id)
-
-      if existing_movie.present?
-        existing_movie
-      else
-        new(
-          title: result[:title],
-          in_db: false,
-          tmdb_id: tmdb_id,
-          release_date: result[:release_date],
-          vote_average: result[:vote_average]&.round(1),
-          overview: result[:overview],
-          backdrop_path: result[:backdrop_path],
-          poster_path: result[:poster_path]
-        )
-      end
+      new(
+        title: result[:title],
+        in_db: false,
+        tmdb_id: tmdb_id = result[:id],
+        release_date: result[:release_date],
+        vote_average: result[:vote_average]&.round(1),
+        overview: result[:overview],
+        backdrop_path: result[:backdrop_path],
+        poster_path: result[:poster_path]
+      )
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/entities/movie_search_result.rb
```diff
@@ -16,25 +16,18 @@ def initialize(title:, in_db:, tmdb_id:, release_date:, vote_average:, overview:
     @lists = nil
   end
 
-  def self.parse_results(results)
+  def self.build_list(results)
```
lortza, line 19: I didn't feel 100% certain that I knew what a `MovieSearch` was, but I know what a `MovieSearchResult` is. Also, `parse` is a little strange, so I changed it to `build_list` since we're actually returning a list of objects.

I feel like the combo of `MovieSearchResult.build_list` is more clear as to what's actually happening here, but I'll open to better words.

app/services/movie_data_service.rb
```diff
@@ -14,6 +14,7 @@ module MovieDataService
   ["unwatched movies", "unwatched movies"], ["only show unwatched", "only show unwatched"],
   ["only show watched", "only show watched"], ["movies not on a list", "movies not on a list"] ]
 
+  # TODO: move these into a TMDB::MenuOptions module
```
lortza, line 17: I believe this is a list of Genres and their TMDB ids. I'd like this to go into some sort of `TMDB` namespaced object in order to make that more clear.

app/views/movies/_movie_partial_loop.html.erb
```diff
@@ -1,9 +1,9 @@
 <div class="row tile-container">
   <% movies.each do |movie| %>
     <article class="tile-cover-pic">
-      <!-- MODAL TRIGGER -->
+      <!-- MODAL TRIGGERS -->
```
lortza, line 4: I tried to make these two triggers more visually obvious since I missed it the first time I looked at this page.

app/views/movies/_movie_partial_loop.html.erb
```diff
@@ -1,24 +1,20 @@
 <div class="row tile-container">
   <% movies.each do |movie| %>
     <article class="tile-cover-pic">
-      <!-- MODAL TRIGGER -->
+      <!-- MODAL TRIGGER 1 of 2 -->
```
lortza, line 4: I tried to make the existence of these 2 different modal triggers more obvious since I missed it the first few times I looked at this file.

app/views/movies/_movie_partial_loop.html.erb
```diff
@@ -1,24 +1,20 @@
...
       </p>
-
+      <!-- MODAL TRIGGER 2 of 2 -->
+      <!-- On Mobile: skip the modal and go straight to a show page -->
       <p id="modal-trigger-mobile">
-        <!-- Get to a show page on mobile regardless of movie object persistence -->
-        <% if movie.is_a?(Movie) %>
-          <%= link_to image_for(movie), movie_path(movie) %>
-        <% else %>
-          <%= link_to image_for(movie), movie_more_path(tmdb_id: movie.tmdb_id) %>
```
lortza, line 20: All `movie`s are instances of `MovieSearchResult`, so we don't need to check. The controller action for the `movie_more_path` is already handling the question of "should i display a persisted movie or hit the api to get fresh data?" with its use of `GuaranteedMovie`.

mikevallano (approved):
Nice!

PR #343, opened 2023-01-04
Add streaming service availability to service provider buttons

### LINKED ISSUE #256: Explore "watch/providers" (streaming availability) endpoint
Not sure how it works, but wanted to make a note to experiment a bit to see if it'd be useful.

https://developers.themoviedb.org/3/movies/get-movie-watch-providers
https://developers.themoviedb.org/3/watch-providers/get-movie-providers

lortza:
## Related Issues & PRs
Closes #256

## Problems Solved
How many clicks are you willing to do to find out if _The Day After Tomorrow_ is free anywhere? Not many. 😆 This PR surfaces the streaming availability as provided by the TMDB API to our existing streaming search convenience buttons. I did not include "Free with Ads" or information from any streaming services that we do not currently use. That means it's pretty hardcoded to us as users and folks with different services won't be able to customize. 

Since we're hitting the API to get this data once per data type (`Movie`, `MovieMore` both respond to a `streaming_service_providers` instance method), I have memoized the responses. Though @mikevallano and I stepped through the code and we haven't uncovered a situation where that memoization is actually being used. So we may want to consider something more cache-conscious for the 1-2 new API calls we're adding with this feature.

## Screenshots

|Before| After | 
|---|---|
|Screen Shot 2023-01-04 at 2 21 10 PM|Screen Shot 2023-01-05 at 6 21 46 PM|

More examples:
| After: with rentals | After: for free | After: buy only |
|---|---|---|
|Screen Shot 2023-01-04 at 2 19 03 PM | Screen Shot 2023-01-04 at 2 19 15 PM| Screen Shot 2023-01-04 at 2 19 35 PM|

### Before
All of the options are present but have know knowledge of actual availability. 

### After
All of our personal streaming services are here displaying pay model information. The information displays the least expensive option for each streaming service provider, for example "**Rent** on Amazon Video" instead of "**Buy** on Amazon Video" when both options are available. If we don't have any streaming availability data for a given provider, the provider is still displayed with "**Try** on..." instead of "Free", "Rent", or Buy". 

## When this Endpoint is Hit
* When clicking the modal for a `Movie` in the db
* By a `movie` when visiting `movies#show`
* When clicking the modal for a `MovieMore` that is not persisted
* By a `movie_more` when visiting `tmdb#movie_more`

## TO DO
- [x] specs for new endpoint
- [x] specs for new views & helpers
- [x] specs for handling when data is missing
- [x] considerations for performance and API efficiency


app/assets/stylesheets/shared.scss
```diff
@@ -106,18 +106,23 @@ color: white !important;
...
-} /* .fa */
+/* FONT AWESOME */
+.fa { line-height: 1.2; }
 
+/* MATERIAL ICONS: https://fonts.google.com/ */
+.material-symbols-outlined {
+  font-size: 1.5rem;
+  position: relative;
+  bottom: -0.2em;
+}
```
lortza, line 125: I couldn't get the FontAwesome icons to work. I think it's a versioning problem. We set up our subscription so long ago. I can dig into it further if you'd like. I just wanted to keep it moving, so I added Material Icons here instead.

app/services/movie_data_service.rb
```diff
@@ -14,7 +14,6 @@ module MovieDataService
   ["unwatched movies", "unwatched movies"], ["only show unwatched", "only show unwatched"],
   ["only show watched", "only show watched"], ["movies not on a list", "movies not on a list"] ]
 
-  # TODO: move these into a TMDB::MenuOptions module
```
lortza, line 17: We talked about this at lunch and decided that we want to let this ride as we don't see a big risk in TMDB updating their genres list in a meaningful way that would break our app.

app/services/movie_data_service.rb
```diff
@@ -250,5 +249,35 @@ def get_common_movies_between_multiple_actors(actor_names: nil, paginate_actor_n
         total_pages: movie_response[:total_pages]
       )
     end
+
+    def get_movie_streaming_service_providers(movie)
+      results = Tmdb::Client.request(:streaming_service_providers, movie_id: movie.tmdb_id).dig(:results, :US)
+      preferred_providers = [
```
lortza, line 255: I got to move the name and url data about each provider out of the view (hardcoded) 😄 and into something more useful here in this object. Even if we weren't doing this fancy API stuff, I'd be tempted to at least do this.

app/services/movie_data_service.rb
```diff
@@ -250,5 +249,35 @@ def get_common_movies_between_multiple_actors(actor_names: nil, paginate_actor_n
...
+        { name: 'Netflix', pay_model: 'try', url: "http://www.netflix.com/search/#{movie.title}"},
+        { name: 'Amazon Prime Video', pay_model: 'try', url: "https://www.amazon.com/s?k=#{movie.title.gsub(' ', '+')}&i=instant-video"},
+        { name: 'Amazon Video', pay_model: 'try', url: "https://www.amazon.com/s?k=#{movie.title.gsub(' ', '+')}&i=instant-video"},
+        { name: 'YouTube', pay_model: 'try', url: "https://www.youtube.com/results?search_query=#{movie.title}+full+movie #{movie.release_date.stamp('2001') if movie.release_date.present?}"},
+        { name: 'Vudu', pay_model: 'try', url: "https://www.vudu.com/content/movies/search?searchString=#{movie.title}"},
+      ]
+      return preferred_providers if results.nil?
```
lortza, line 262: In the case that the API gives us absolutely no results, we'll populate the page with our default services as "Try"s (ex: "Try on Netflix") which is the same functionality we have now.

app/services/movie_data_service.rb
```diff
@@ -250,5 +249,35 @@ def get_common_movies_between_multiple_actors(actor_names: nil, paginate_actor_n
...
+
+      free_provider_names = results[:flatrate]&.map { |result| result[:provider_name] } || []
+      rent_provider_names = results[:rent]&.map { |result| result[:provider_name] } || []
+      buy_provider_names = results[:buy]&.map { |result| result[:provider_name] } || []
+
+      preferred_providers.map do |provider|
```
lortza, line 268: The order of this conditional block is important. It ensures that we're displaying the least-expensive option for each provider. If a provider does not appear in any of the payment model results, it is assigned a "try", which translates in the view as "go look it up, we don't know".

spec/features/movies_feature_spec.rb
```diff
@@ -146,10 +150,14 @@
 
       end #movie is on a list
 
+      scenario "the user can see streaming service options" do
+        sign_in_user(user)
+        visit(movie_path(movie))
+        expect(page).to have_content("FakeFlix")
+      end
```
lortza, line 157: This spec confirms that we actually see this data on the page.

spec/features/movies_feature_spec.rb
```diff
@@ -230,7 +237,11 @@
 
       end #tagging context
 
-      context "pagingation" do
+      context "pagination" do
+        before do
+          allow(MovieDataService).to receive(:get_movie_streaming_service_providers)
+                                 .and_return(streaming_service_providers)
```
lortza, line 243: The rest of these `allow`s just satisfy the page's need for data.

spec/features/movies_feature_spec.rb
```diff
@@ -246,12 +257,6 @@
             create(:tagging, movie_id: counter, user_id: user.id, tag_id: tag.id)
             counter += 1
           end
-          # visit root_path
-          # visit movies_path
-          # @movie = Movie.first
-          # find("#modal_link_#{@movie.tmdb_id}")
-          # find("#modal_link_#{@movie.tmdb_id}").click
-          # click_link "hilarious"
```
lortza, line 254: If we find we need this, we can put it back.

mikevallano: yeah, let's keep it because we have an issue to fix up the failing feature specs

spec/services/movie_data_service_spec.rb
```diff
@@ -393,4 +393,244 @@
...
+              }
+            }
+            allow(Tmdb::Client).to receive(:request).and_return(response_data)
+            results = MovieDataService.get_movie_streaming_service_providers(movie)
+
+            expect(results.find {|r| r[:name] == 'Vudu'}[:pay_model] ).to eq('rent')
+        end
+      end
+
+      describe 'when multiple preferred providers appear in multiple categories' do
```
mikevallano, line 579: nice tests 😎

spec/services/movie_data_service_spec.rb
```diff
@@ -393,4 +393,244 @@
...
+                rent:
+                  [{:logo_path=>"/peURlL.jpg", :provider_id=>2, :provider_name=>"Vudu", :display_priority=>4},
+                  {:logo_path=>"/5NyLm4.jpg", :provider_id=>10, :provider_name=>"Amazon Video", :display_priority=>12}],
+              }
+            }
+          }
+          allow(Tmdb::Client).to receive(:request).and_return(response_data)
+          results = MovieDataService.get_movie_streaming_service_providers(movie)
+
+          aggregate_failures "expected pay_model levels per provider" do
```
mikevallano, line 600: sweet!

mikevallano (approved):
awww yeah! 🚀 🎉

Issue #307, opened 2022-03-14
Make autocompletes more efficient


lortza:
Right now, we populate our 3 autocomplete fields with just the `name` of the data (movie, person, tv series). We then take that name and pass it straight to a subsequent API call where we need to get the id. It would make more sense to return both the name and id for the autocomplete field and then submit the id instead of the name.

lortza:
Also, like we discussed today, we should not be making a request with every keystroke. We should add a minimum character count and then reconsider a single keystroke.

mikevallano:
Here are a couple options for jquery autocomplete: https://stackoverflow.com/a/41347730

PR #345, opened 2023-01-06
Add delay to search autocomplete 

lortza:
## Related Issues & PRs
Relates to #307

## Problems Solved
While we do have amazing responsiveness in our autocomplete fields, we're feeling a little jerkish for hammering on the free TMDB API. This PR adds a delay to the call to the API that happens during autocomplete -- in effect, reducing the calls to the API by about half without too much of a wait on the user's part.

## Screenshots
Before
before2

After
after


mikevallano (approved):
🚀 Awesome! big impact with little code!

PR #349, opened 2023-01-09
Add last-watched date and count to seen-movies section of actor bio

### LINKED ISSUE #313: Add more details to "Movies I've Seen"
Now that I can see the "movies i've seen" for a given actor, i'd like a couple of more details. Here, i can see that i've seen a bunch of Kevin Hart movies. 
Screen Shot 2022-06-15 at 6 01 35 PM

But _how many_ is that? I'm lazy and don't want to count. I'd like to have a count ( `MOVIES I'VE SEEN (15)`) at the top like we do here in the section above:
Screen Shot 2022-06-15 at 6 03 40 PM

 And it would also be neat to incorporate my rating and "last seen" information. maybe 🤷 
Screen Shot 2022-06-15 at 6 04 04 PM

lortza:
## Related Issues & PRs
Closes #313

## Problems Solved
This PR adds more data to the seen-movies section of the actor bio. This way, we'll have more context when looking at that list of movies and know the answer to the questions:
* when did I see this movie last?
* how many movies have I seen with this actor in it? 

## Screenshots
| Before | After |
|---|---|
|  Screen Shot 2023-01-09 at 2 10 28 PM |Screen Shot 2023-01-09 at 2 09 22 PM |

## Performance Concerns
Getting this new data was not particularly costly. For the count of movies I've sen, it's just a `.length` on an array. For adding the last-watched date, I'm adding a db lookup for each movie. This could be costly, but the reality of it is that this dataset will always be small. I doubt either of us will ever have seen any more than 20 movies that a specific actor has been in. So it really doesn't concern me. I started looking into some  fancier querying, but since I also had to consider the data coming from the API, it just seemed to get complicated and the juice didn't seem worth the squeeze.

### DIFF
--- a/app/views/tmdb/actor_more.html.erb
+++ b/app/views/tmdb/actor_more.html.erb
@@ -14,7 +14,7 @@
     <div class="credits">
       <%= render partial: "movie_credits", locals: { person_profile: @data.actor.profile, movie_credits: @data.actor.movie_credits } %>
 
-      <%= render partial: "seen_movie_credits", locals: { person_profile: @data.actor.profile, movies_seen: @data.movies_seen } %>
+      <%= render partial: "seen_movie_credits", locals: { movies_seen: @data.movies_seen } %>
 
       <%= render partial: "tv_credits", locals: { tv_credits: @data.actor.tv_credits, actor_id: @data.actor.person_id }  %>
 
--- a/app/controllers/tmdb_controller.rb
+++ b/app/controllers/tmdb_controller.rb
@@ -76,14 +76,38 @@ class TmdbController < ApplicationController
     end
   end
 
+  # def actor_more
+  #   actor_data = PersonDataService.get_person_profile_data(params[:actor_id])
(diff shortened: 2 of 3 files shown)


app/views/tmdb/actor_more.html.erb
```diff
@@ -14,7 +14,7 @@
     <div class="credits">
       <%= render partial: "movie_credits", locals: { person_profile: @data.actor.profile, movie_credits: @data.actor.movie_credits } %>
 
-      <%= render partial: "seen_movie_credits", locals: { person_profile: @data.actor.profile, movies_seen: @data.movies_seen } %>
```
lortza, line 17: These details are folded into the credits data now so this extra local variable is not needed

mikevallano (approved):
🚀 Another sweet featch! 

I created a separate ticket for myself to clean up queries: https://github.com/mikevallano/tmdb-moviequeue/issues/350

Issue #354, opened 2023-01-13
Revisit sorting and filtering to what we actually want


mikevallano:
In refactoring queries, it makes us question if the current sorting/filtering makes the most sense.

We should remove any sorting we aren't using, and add anything we want.

PR #355, opened 2023-01-14
350: Clean up model queries to use SQL over ruby

### LINKED ISSUE #350: Refactor queries in models to use correct SQL, and not use ruby
There are some poor queries, primarily in the movies model.

I'd like to clean them up.

Also, it'd be good to ensure we aren't doing an N+1 here: https://github.com/mikevallano/tmdb-moviequeue/pull/349/files

This is the first step in refactoring queries.

This ticket will also require adding user model specs.

mikevallano:
## Related Issues & PRs
Closes #350

## Description
This PR is a refactor to clean up model queries to use sql instead of ruby and shouldn't change any behavior (aside from potentially fixing some sorts that weren't working correctly).
It sets us up to revisit the sorting we actually want: https://github.com/mikevallano/tmdb-moviequeue/issues/354 

## Problems Solved
* Uses sql instead of ruby for sorting methods
* Uses better sql syntax where needed
* Adds specs to ensure sorting methods are working correctly
* Fixes N+1 query in the actor show page

## Screenshots
https://user-images.githubusercontent.com/7945260/213922098-3aabd989-a0d7-4b83-96eb-e2b509b89154.mov

## Things Learned
* Some of the queries were pretty tricky. I learned about `NULLS FIRST` and `NULLS LAST` in postgres: https://stackoverflow.com/a/9511492
* We had poor test coverage in the users model
* There's lots more to clean up in the sorting handler, etc. 😅


app/models/movie.rb
```diff
@@ -21,40 +21,57 @@ class Movie < ApplicationRecord
...
+      LEFT OUTER JOIN screenings
+        ON screenings.movie_id = movies.id
+        AND screenings.user_id = #{user.id}
+    SQL
+    select('movies.*, max(screenings.date_watched) AS max_date_watched')
+      .joins(join_query)
+      .group('movies.id')
   end
 
-  def self.by_watched_by_members(list)
```
mikevallano, line 48: This is the query that wasn't being used anywhere, so I removed it and its associated spec.

spec/features/movies_feature_spec.rb
```diff
@@ -25,420 +25,417 @@
       { name: "Foodoo", url: "https://www.foodoo.com/search?searchString=Fake", pay_model: "rent" }
     ]}
 
-    describe "movie show page functionality" do
+    describe 'move show page' do
       before do
-        allow(StreamingServiceProviderDataService).to receive(:get_providers).and_return(streaming_service_providers)
+        allow(StreamingServiceProviderDataService)
+          .to receive(:get_providers)
+          .and_return(streaming_service_providers)
```
mikevallano, line 32: Really the only change in this file is to add one parent describe block so that the stubbing of StreamingServiceProviderDataService only needs to happen once. Prolly best to view with whitespace off.

spec/features/movies_feature_spec.rb
```diff
@@ -25,420 +25,417 @@
...
+        scenario "movie show page has the movie's poster image" do
+          sign_in_user(user)
+          listing
           visit(movie_path(movie))
-          fill_in "tag_list", with: "dark comedy"
-          click_button "add_tags_button_movies_partial", match: :first
-          expect(page).to have_content("dark-comedy")
-          click_link "remove_tag_link_movies_partial"
-          expect(page).not_to have_content("dark-comedy")
+          expect(page).to have_css("img[src*='https://image.tmdb.org/t/p/w185#{movie.poster_path}']")
```
mikevallano, line 63: This is the only other change in this file, and it's just to make it less brittle since I changed the way the factory handles the image.

spec/models/movie_spec.rb
```diff
@@ -110,39 +110,35 @@
         membership1; membership2
       end
 
-      it 'sorts by watched by user' do
-        expect(Movie.by_watched_by_user(list, user).first).to eq(fargo)
-      end
-
-      it 'sorts by watched by members' do
```
mikevallano, line 117: This is the spec for the method that wasn't being used.

spec/models/user_spec.rb
```diff
@@ -1,113 +1,129 @@
...
-  it { is_expected.to have_many(:member_movies) }
-  it { is_expected.to have_many(:tags) }
-  it { is_expected.to have_many(:taggings) }
-  it { is_expected.to have_many(:sent_invites) }
-  it { is_expected.to have_many(:reviews) }
-  it { is_expected.to have_many(:reviewed_movies) }
-  it { is_expected.to have_many(:ratings) }
-  it { is_expected.to have_many(:rated_movies) }
-  it { is_expected.to have_many(:screenings) }
-  it { is_expected.to have_many(:watched_movies) }
```
mikevallano, line 30: I removed these relation specs because they're pretty pointless and just take up space.

spec/models/user_spec.rb
```diff
@@ -1,113 +1,129 @@
...
 
 RSpec.describe User, type: :model do
-  let(:user) { FactoryBot.create(:user) }
-  let(:invalid_user) { FactoryBot.build(:invalid_user) }
-  let(:list) { FactoryBot.create(:list, owner_id: user.id) }
-  let(:list2) { FactoryBot.create(:list, owner_id: user.id) }
-  let(:movie) { FactoryBot.create(:movie) }
-  let(:movie2) { FactoryBot.create(:movie) }
-  let(:listing) { FactoryBot.create(:listing, list_id: list.id, user_id: user.id, movie_id: movie.id ) }
-  let(:listing2) { FactoryBot.create(:listing, list_id: list2.id, user_id: user.id, movie_id: movie2.id ) }
```
mikevallano, line 11: These were updated to not use `FactoryBot` since it's not needed.

spec/rails_helper.rb
```diff
@@ -46,7 +46,9 @@
   # Capybara::Selenium::Driver.new(app, browser: :chrome, desired_capabilities: capabilities)
 end
 
-# use selenium_chrome to actually see what's doing
+# To actually see the tests in the browser locally,
+# use selenium_chrome instead of selenium_chrome_headless
+# Capybara.javascript_driver = :selenium_chrome
```
mikevallano, line 51: Just added a commented out example to make it easier to not have to remember the syntax when we want to run the feature tests in chrome locally.

app/models/movie.rb
```diff
@@ -21,40 +21,57 @@ class Movie < ApplicationRecord
...
-  scope :by_shortest_runtime, -> { order(:runtime) }
-  scope :by_longest_runtime, -> { order(:runtime).reverse_order }
-  scope :by_recent_release_date, -> { order(:release_date).reverse_order }
-  scope :by_highest_vote_average, -> { order(:vote_average).reverse_order }
+  scope :by_shortest_runtime, -> { order(runtime: :asc) }
+  scope :by_longest_runtime, -> { order(runtime: :desc) }
+  scope :by_recent_release_date, -> { order(release_date: :desc) }
+  scope :by_highest_vote_average, -> { order(vote_average: :desc) }
   scope :default_list_order, -> (list) do
-    list.movies.order('CAST(listings.created_at as DATE) desc, listings.priority desc')
```
mikevallano, line 29: This should fix that deprecation warning you mentioned, but I'm not sure how to test it.

app/models/movie.rb
```diff
@@ -21,40 +21,57 @@ class Movie < ApplicationRecord
...
+  def self.screenings_join_query(user)
+    join_query = <<~SQL.squish
+      LEFT OUTER JOIN screenings
+        ON screenings.movie_id = movies.id
+        AND screenings.user_id = #{user.id}
+    SQL
+    select('movies.*, max(screenings.date_watched) AS max_date_watched')
+      .joins(join_query)
+      .group('movies.id')
   end
```
mikevallano, line 53: This is the main query used by a few methods so I moved it to its own method.

It allows sorting by unwatched/watched movies, etc. 

It took me a bit to figure it out 😅 .

Basically, left outer joining by movie _and_ user will only populate results if the user has seen the movie.

Grouping by `movies.id` ensures we get distinct results (adding a distinct won't work because of the left outer join).

app/models/user.rb
```diff
@@ -49,71 +49,92 @@ def should_generate_new_friendly_id?
   source: :movie
 
   def all_lists
-    (self.lists | self.member_lists).uniq
+    list_ids = lists.pluck(:id)
+    member_list_ids = member_lists.pluck(:id)
+    ids = (list_ids + member_list_ids).uniq
```
mikevallano, line 54: This query, `all_listings` and `all_movies` use multiple queries.

Given the way we have tables set up, I don't know of any other way to accomplish it without making a couple queries.

Even if there is some way to do it, it's probably more efficient to do a couple simple queries than a huge query that joins several tables.

app/models/user.rb
```diff
@@ -49,71 +49,92 @@ def should_generate_new_friendly_id?
...
+    return all_lists_by_name unless movie&.in_db
+    movie_list_ids = movie.lists.by_user(self).pluck(:id)
+    all_lists.where.not(id: movie_list_ids)
   end
 
   def all_lists_by_name
-    self.all_lists.sort_by { |list| list.name.downcase }
+    # TODO: not sure what this is acting weird
+    # all_lists.order('lower(name) ASC')
+    all_lists.sort_by { |list| list.name.downcase }
```
mikevallano, line 74: This one I did not update to use sql because the results were fucky. I need to look more into it, but for now, I just kept it as-is.

app/controllers/tmdb_controller.rb
```diff
@@ -79,7 +79,11 @@ def two_actor_search
   def actor_more
     actor_data = PersonDataService.get_person_profile_data(params[:actor_id])
     actor_movie_ids = actor_data.movie_credits.actor.map { |m| m.tmdb_id }
-    user_movies_seen = current_user.watched_movies.where(tmdb_id: actor_movie_ids).uniq
+    user_movies_seen =
+      current_user
+      .watched_movies_with_max_screening_date
+      .where(tmdb_id: actor_movie_ids)
+      .distinct
```
mikevallano, line 86: This is a way to get the data without an N+1 🎉

app/controllers/tmdb_controller.rb
```diff
@@ -89,7 +93,7 @@ def actor_more
         title: movie.title,
         character: credit.character,
         birthday: actor_data.profile.birthday,
-        last_watched_date: movie.most_recent_screening_by(current_user)
+        last_watched_date: movie.max_screening_date
```
mikevallano, line 96: Since we're selecting `max(screenings.date_watched)` and giving it the alias in the query, we have access to the alias here.

app/models/movie.rb
lortza: 🤓 nice work! This all looks so much better now too! 💪

app/models/user.rb
```diff
@@ -49,71 +49,92 @@ def should_generate_new_friendly_id?
...
+    list_movie_ids = movies.pluck(:id)
+    member_movie_ids = member_movies.pluck(:id)
+    ids = (list_movie_ids + member_movie_ids).uniq
+    Movie.where(id: ids).by_genre(genre)
+  end
+
+  def watched_movies_with_max_screening_date
+    watched_movies
+      .select("movies.*, MAX(TO_CHAR(screenings.date_watched, 'mm/dd/yyyy')) AS max_screening_date")
+      .group('movies.id')
```
lortza, line 137: awwwww yyyyeaaahhhh! these queries are sooo much better now!

spec/factories/movies.rb
```diff
@@ -4,7 +4,9 @@
     sequence(:tmdb_id) { |n| n + 10 }
     imdb_id { FFaker::Product.model }
     backdrop_path { '/lGAaaOzqw8nc14HOgSP58TWWo1y.jpg' }
-    poster_path { '/aZeX4XNSqa08TdMHRB1gDLO6GOi.jpg' }
+    # this will lead to a broken link, but it helps with feature specs
+    # to know which movie it is, rather than hard-coding a specific movie
+    poster_path { "#{title}.jpg" }
```
lortza, line 9: it's time to turn off the gas light, eh? 😆 😢

spec/features/lists_feature_spec.rb
```diff
@@ -185,12 +194,12 @@
...
       scenario "user can update a listing's priority", js: true do
-        listing
+        create(:listing, list: list, movie: fargo)
         page.driver.browser.manage.window.resize_to(1280,800)
         sign_in_user(user)
         visit(user_list_path(user, list))
-        find("#modal_link_#{movie.tmdb_id}")
-        find("#modal_link_#{movie.tmdb_id}").click
+        find("#modal_link_#{fargo.tmdb_id}")
+        find("#modal_link_#{fargo.tmdb_id}").click
```
lortza, line 202: Good clarification here -- especially given your last struggs with all of that hardcoded stuff leading to confusion. 😓

.gitignore
```diff
@@ -30,6 +30,7 @@ tester2.rb
 tester.html.erb
 tester2.html.erb
 tester.js.erb
+misc.sql
```
lortza, line 33: What is this?

lortza (approved):
This all looks good to me. You did so much great cleanup of queries and tests too 😍 🚀 🥇

.gitignore
mikevallano: just a scratch pad. we should just move this stuff to a tmp folder and exclude that

PR #357, opened 2023-01-17
356 Add watch providers for TV shows

### LINKED ISSUE #356: Add screening service providers for tv shows
I'd like to add streaming service providers to TV shows just like we have for movies. It should be a fairly easy lift since all of the hard part of parsing and displaying was handled in the movies version. #343 

Also, we need to add watch-provider data credit somewhere. When I added our sweet new streaming service info, I failed to follow the rules set forth by the generous folks at TMDB. We need to credit JustWatch somewhere.

> Powered by our partnership with JustWatch, you can query this method to get a list of the availabilities per country by provider.
>
> This is not going to return full deep links, but rather, it's just enough information to display what's available where.
>
> You can link to the provided TMDB URL to help support TMDB and provide the actual deep links to the content.
> 
> Please note: In order to use this data you must attribute the source of the data as JustWatch. If we find any usage not complying with these terms we will revoke access to the API.

https://developers.themoviedb.org/3/tv/get-tv-watch-providers

lortza:
## Related Issues & PRs
Closes #356

This PR does 3 things for us regarding streaming service providers:
1. adds service provider info to the TV Series page
1. fulfills our obligation to TMDB by crediting [JustWatch](https://www.justwatch.com/) for this data
1. adds Hoopla to our list of curated service providers

In the process, I refactored `MovieDataService.get_movie_streaming_service_providers` into a new object `StreamingServiceProviderDataService.get_providers` that handles watch provider data for both movies and tv. 

## Screenshots
This is an example of a the provider data for "The Forgiven" that we watched this weekend via Hoopla. Yay! Another free movie option.
Screen Shot 2023-01-17 at 7 31 39 PM

And this is the TV Series page with hot new data in it. I've also added the sidebar to house this data much like the way it appears in the movies iteration of this feature.
Screen Shot 2023-01-17 at 9 59 34 PM


app/services/movie_data_service.rb
```diff
@@ -249,56 +249,5 @@ def get_common_movies_between_multiple_actors(actor_names: nil, paginate_actor_n
         total_pages: movie_response[:total_pages]
       )
     end
-
-    def get_movie_streaming_service_providers(movie)
```
lortza, line 253: This is replaced by a slightly more robust version that can also handle requests for `tv` data.

app/services/streaming_service_provider_data_service.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module StreamingServiceProviderDataService
+  class << self
+    def get_providers(tmdb_id:, title:, media_type:, media_format:, release_date: nil)
```
lortza, line 5: While at first it felt a little obnoxious going from one argument to 4-5 keyword arguments, this approach makes the method much more flexible while divesting itself from any knowledge it was holding about the shape of a `movie` object. 😄  Now it just takes in the data it needs and spits out results.

app/views/shared/_streaming_service_providers.html.erb
```diff
@@ -5,4 +5,5 @@
        <%= provider[:pay_model].titleize %> on <%= provider[:name] %>
     </a>
   </p>
-<% end %>
\ No newline at end of file
+<% end %>
+<p class='txt-sm'>Availability data by <%= link_to 'JustWatch', 'https://www.justwatch.com', target: '_blank' %></p>
```
lortza, line 9: This fulfills our obligation to TMDB.

app/services/streaming_service_provider_data_service.rb
```diff
@@ -0,0 +1,66 @@
...
+      params = { tmdb_id: tmdb_id, media_type: media_type }
+      results = Tmdb::Client.request(:streaming_service_providers, params)&.dig(:results, :US)
+
+      parameterized_title = ActiveSupport::Inflector.parameterize(title, separator: '+')
```
lortza, line 9: This calling of `ActiveSupport::Inflector` feels kind of janky to me. Do you have any suggestions?

app/services/streaming_service_provider_data_service.rb
```diff
@@ -0,0 +1,66 @@
...
+      default_providers = [
+        # Find more providers at TMDB's watch_provider partner: https://www.justwatch.com/us
+        {
+          name: 'Hoopla',
```
lortza, line 13: Oh snap! Now we get free library movies! 💃

app/views/tmdb/tv_series.html.erb
```diff
@@ -1,40 +1,49 @@
 <article id="profile">
```
lortza, line 1: I recommend turning off whitespaces in the github diff view to see this file best.

lib/tmdb/client.rb
```diff
@@ -35,7 +35,7 @@ def build_url(endpoint, params)
         when :tv_season_data then "/tv/#{params[:series_id]}/season/#{params[:season_number]}?api_key=#{API_KEY}&append_to_response=credits"
         when :tv_episode_data then "/tv/#{params[:series_id]}/season/#{params[:season_number]}/episode/#{params[:episode_number]}?api_key=#{API_KEY}"
         when :multi_search then "/search/multi?api_key=#{API_KEY}&query=#{searchable_query(params[:query])}"
-        when :streaming_service_providers then "/movie/#{params[:movie_id]}/watch/providers?api_key=#{API_KEY}"
+        when :streaming_service_providers then "/#{params[:media_type]}/#{params[:tmdb_id]}/watch/providers?api_key=#{API_KEY}&watch_region=US"
```
lortza, line 38: This `watch_region` argument does not seem to work. 🤔 I'm thinking I should probably remove it. What do you think?

app/services/streaming_service_provider_data_service.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module StreamingServiceProviderDataService
+  class << self
+    def get_providers(tmdb_id:, title:, media_type:, media_format:, release_date: nil)
+      params = { tmdb_id: tmdb_id, media_type: media_type }
+      results = Tmdb::Client.request(:streaming_service_providers, params)&.dig(:results, :US)
+
+      parameterized_title = title.parameterize(separator: '+')
```
lortza, line 9: This evening when I was annoyed with this [`ActiveSupport::Inflector.parameterize` method](https://apidock.com/rails/ActiveSupport/Inflector/parameterize), I did a little more digging and found the [`String` class's version](https://apidock.com/rails/v5.2.3/String/parameterize) which actually does allow me to pass a separator. 😄

mikevallano: nice! 😎

app/services/streaming_service_provider_data_service.rb
mikevallano: daaaaaaaaaamn!

lib/tmdb/client.rb
mikevallano: 🤔 I'll defer to you since I haven't looked at the api in quite a while and I'm not sure how you are determining whether it works or not. I'm happy to discuss, but off hand, dunno.

spec/services/streaming_service_provider_data_service_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+        end
+      end
+    end
+
+    describe 'when the API does return results' do
+      describe 'but the results do not have free, flatrate, rent, or buy pay model options' do
+        it 'returns our default providers with a pay_model of "try"' do
+          response_data = {
+            results: {
+              US: {
```
mikevallano, line 39: not sure if this is related to the `watch_region question above.

spec/services/streaming_service_provider_data_service_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+                ads:
+                  [{:logo_path=>"/5OAb2w.jpg", :provider_id=>634, :provider_name=>"Netflix", :display_priority=>26},
+                  {:logo_path=>"/eWp5Ld.jpg", :provider_id=>43, :provider_name=>"Amazon Video", :display_priority=>38},
+                  {:logo_path=>"/xL9SUR.jpg", :provider_id=>358, :provider_name=>"YouTube", :display_priority=>51}],
+                subscribe:
+                  [{:logo_path=>"/peURlL.jpg", :provider_id=>2, :provider_name=>"Vudu", :display_priority=>4},
+                  {:logo_path=>"/peURlL.jpg", :provider_id=>2, :provider_name=>"Amazon Prime Video", :display_priority=>4}],
```
mikevallano, line 46: i'm sure you got this from copying the actual api response, but we should add rubocop at some point soon so that it would automatically fix these hashrockets.

i wonder if there some example starter rubocop setups out there we could grab so it wouldn't be a huge undertaking?

mikevallano (approved):
🎉 🚀 Awwww yeah! 

I'll defer to you on the question you had, but otherwise looks great!

spec/services/streaming_service_provider_data_service_spec.rb
lortza: I know there is a way to only run on modified files ([here is one post about it](https://dev.to/risafj/running-rubocop-only-on-modified-files-in-a-project-with-no-rubocop-2ioe)). I'm wondering if there is a way to configure rubocop itself to run on modified only. 🤔  That's something to look into for another PR.

spec/services/streaming_service_provider_data_service_spec.rb
lortza: Sort of. I assume if they filtered out the results by our chosen region for us, I wouldn't have to dig for it and it wouldn't look like this in the specs. Alas, their filtering does not work, so I dig for it.

spec/services/streaming_service_provider_data_service_spec.rb
mikevallano: Neat. Okay, I created an issue to investigate some sort of rubocop solution: https://github.com/mikevallano/tmdb-moviequeue/issues/358

PR #365, opened 2023-01-21
Make media profile layout that's shared across Movie, MovieMore, TVSeries

lortza:
## Related Issues & PRs
Closes #364 

## Problems Solved
We recently implemented displaying streaming service providers for movies, making the TV and Movie/MovieMore profiles even more similar. This PR streamlines the code between those different classes, making it easier to maintain features across all 3 classes. 

I originally wanted to do a CSS refactor so that classes weren't called things like `movie-show` and to incorporate a little BEM into our conventions. But it got hairy. I decided to stick to the task at hand and just clean up the views. In the process, I removed (perhaps the last?ish?) of the `@movie` instance variables and converted them to locals. 

## Things Learned
* CSS gets hairy. It can be hard to untangle. It's worthy of its own refactor PR.
* Media queries and browser profiles don't necessarily match actual device settings. I had a real struggle trying to get styles to render as expected on the tablet even after I made a browser profile specifically for our tablet. 🤷 This is a fight for another day.

PR #366, opened 2023-01-30
Resolve blank form sentry errors

lortza:
## Related Issues & PRs
* [TMDB-MOVIEQUEUE-6Z](https://mike-vallano.sentry.io/issues/3901921697/?project=127874&query=is%3Aunresolved&referrer=issue-stream)
* [TMDB-MOVIEQUEUE-70](https://mike-vallano.sentry.io/issues/3904770136/?project=127874&query=is%3Aunresolved&referrer=issue-stream)

## Problems Solved
1. We were getting an undefined variable error in the `movies#show` action because I missed switching it over from `movie` to `@media` in my last PR #365.

2. We discovered that we did not have any error handling for when a user submits an empty string to any of the searches:
  * movie by title
  * actor by name
  * tv series by name
  * movies between two actors
  * actors between two movies
  To resolve this, I added handling and specs for the handling of empty strings at the `Tmdb::Client` level. I also added `required: true` to all of the form fields that are indeed required. This prevents us from even getting to the error handling in the `Tmdb::Client`.

## Screenshots
Now all of the form fields have this little error text (courtesy of default HTML behavior) if a user submits an empty form.
Screen Shot 2023-01-30 at 10 15 28 AM

### DIFF
--- a/app/controllers/movies_controller.rb
+++ b/app/controllers/movies_controller.rb
@@ -38,7 +38,7 @@ class MoviesController < ApplicationController
 
   def show
     @media = Movie.friendly.find(params[:id])
-    return redirect_to movie, status: :moved_permanently if request.path != movie_path(@media)
+    return redirect_to movie_path(@media), status: :moved_permanently if request.path != movie_path(@media)
 
     render 'shared/media_profile'
   end
--- a/app/views/tmdb/actor_search.html.erb
+++ b/app/views/tmdb/actor_search.html.erb
@@ -3,8 +3,17 @@
 <h1>Search by Actor</h1>
 <% if @results&.actor.blank? %> <!-- Display the search form -->
   <%= form_tag '/tmdb/actor_search', { class: "form-class", id: "actor-search", role: "form", method: :get } do %>
-    <%= text_field_tag :actor, nil, class: "form-control search-form autocomplete-search-field", id: "actor_name_actor_search", placeholder: "Enter Actor Name", data: { autocomplete_source: person_autocomplete_path } %>
-    <%= select_tag :sort_by, options_for_select(MovieDataService::SORT_BY), {:prompt => "Sort by...", class: "form-control search-form", :include_blank => false } %>
+    <%= text_field_tag :actor, nil,
+      class: "form-control search-form autocomplete-search-field",
+      id: "actor_name_actor_search",
+      required: true,
+      placeholder: "Enter Actor Name",
+      data: { autocomplete_source: person_autocomplete_path }
+    %>
+    <%= select_tag :sort_by,
+      options_for_select(MovieDataService::SORT_BY),
+      { prompt: "Sort by...", class: "form-control search-form", include_blank: false }
+    %>
     <%= submit_tag "Search", id: "submit_button_actor_search", class: "form-control-submit search-form-submit" %>
   <% end %>
 <% end %>
--- a/app/views/tmdb/search.html.erb
+++ b/app/views/tmdb/search.html.erb
@@ -1,29 +1,37 @@
 <% content_for(:title, "Search by Movie Title") %>
 
-<% if !@movie_title.present? %>
+<% if @movie_title.blank? %>
   <h1 id="search_by_title_header">Search for a Movie by Title</h1>
-  <%= form_tag '/tmdb/search', { class: "form-class", id: "movie-title-search", role: "form", method: :get } do %>
-    <%= text_field_tag :movie_title, nil, class: "form-control search-form search-form autocomplete-auto-submit", id: "movie_field_movie_search", placeholder: "Enter Movie Title", data: { autocomplete_source: movie_autocomplete_path }  %>
+  <%= form_tag '/tmdb/search', { class: "form-class mb-10", id: "movie-title-search", role: "form", method: :get } do %>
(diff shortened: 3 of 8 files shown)


app/controllers/movies_controller.rb
```diff
@@ -38,7 +38,7 @@ def update
 
   def show
     @media = Movie.friendly.find(params[:id])
-    return redirect_to movie, status: :moved_permanently if request.path != movie_path(@media)
```
lortza, line 41: This resolves [TMDB-MOVIEQUEUE-70](https://mike-vallano.sentry.io/issues/3904770136/?project=127874&query=is%3Aunresolved&referrer=issue-stream)

app/views/tmdb/search.html.erb
```diff
@@ -1,29 +1,37 @@
 <% content_for(:title, "Search by Movie Title") %>
 
-<% if !@movie_title.present? %>
+<% if @movie_title.blank? %>
```
lortza, line 3: I prefer to use a positive expression instead of a double negative when possible.

app/views/tmdb/actor_search.html.erb
```diff
@@ -3,8 +3,17 @@
 <h1>Search by Actor</h1>
 <% if @results&.actor.blank? %> <!-- Display the search form -->
   <%= form_tag '/tmdb/actor_search', { class: "form-class", id: "actor-search", role: "form", method: :get } do %>
-    <%= text_field_tag :actor, nil, class: "form-control search-form autocomplete-search-field", id: "actor_name_actor_search", placeholder: "Enter Actor Name", data: { autocomplete_source: person_autocomplete_path } %>
-    <%= select_tag :sort_by, options_for_select(MovieDataService::SORT_BY), {:prompt => "Sort by...", class: "form-control search-form", :include_blank => false } %>
+    <%= text_field_tag :actor, nil,
+      class: "form-control search-form autocomplete-search-field",
+      id: "actor_name_actor_search",
+      required: true,
```
lortza, line 9: Here I've added `required: true` and broken the `text_field_tag` onto multiple lines. This is the same pattern I've done in all 5 search views.

lib/tmdb/client.rb
```diff
@@ -11,6 +11,8 @@ class Error < StandardError; end
 
     class << self
       def request(endpoint, params)
+        return { results: [] } if params.blank?
+        return { results: [] } if params[:query] == ''
```
lortza, line 15: This is the client logic to resolve [TMDB-MOVIEQUEUE-6Z](https://mike-vallano.sentry.io/issues/3901921697/?project=127874&query=is%3Aunresolved&referrer=issue-stream) error when a user submits an empty string.

lib/tmdb/client.rb
```diff
@@ -11,6 +11,8 @@ class Error < StandardError; end
...
         return { results: [] } if params[:query].present? && searchable_query(params[:query]).empty?
```
lortza, line 16: This is the client logic to resolve [TMDB-MOVIEQUEUE-6Z](https://mike-vallano.sentry.io/issues/3901921697/?project=127874&query=is%3Aunresolved&referrer=issue-stream) error when a user submits an empty string. It's kind of weird to have 3 lines that all return the same thing. Any ideas on how to make this prettier and still easy to understand?

mikevallano: Since `searchable_query` returns nil if query isn't present, I think we could just do one line with:

`return { results: [] } if params.blank? || searchable_query(params[:query]).blank?`

app/views/tmdb/actor_search.html.erb
mikevallano: nice! 🚀

app/views/tmdb/search.html.erb
mikevallano: totes!

mikevallano (commented):
looks great! I just had the one comment about the multiple lines thing you asked about.

lib/tmdb/client.rb
lortza: I'm not sure that does everything we need it to. It may the the case that the params are blank -- which would look like `{}`. It may be a separate case where the params aren't blank, but the specific value for the keyword `query` is blank -- which would look like `{ query: '' }`. The third case we're guarding against is if the `params[:query]` is actually full of symbols, which are not searchable.

mikevallano: well, i tried it and your tests passed 😆 . 

i guess we'd need a test case for the scenarios you're talking about, and then we can tweak the logic to see if the tests pass 🤷

mikevallano: to clarify what we talked about, I pulled this branch and updated the code to the one line I suggested above, then ran the tests for the file on this branch, and they passed. 

so it seems like either the one line suggested above would work, or we need additional test cases 😅

lib/tmdb/client.rb
```diff
@@ -59,8 +59,15 @@ def build_url_for_movie_discover_search(params)
         api_path
       end
 
+      def invalid_params?(params)
```
mikevallano, line 62: 🎉 nice!

mikevallano (approved):
Looks great! Nice work adding the additional tests too!

PR #368, opened 2023-02-02
367: Remove coveralls

### LINKED ISSUE #367: Remove coveralls
They are no longer offering a free tier, and there's really no point in using it in this repo

mikevallano:
## Related Issues & PRs
Closes #367 

## Problems Solved
* Just removes coveralls, which does an analysis and tells you how much of your code has test coverage. It was saying we have like 34% or something, which is not accurate.

Anyway, they stopped offering their free tier, so I thought we should just remove it

## Screenshots
Before:
coveralls-before
After:
coveralls-after

## Things Learned
* I learned I don't know how coveralls works, because we have way better than 34% coverage

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,6 @@ gem 'friendly_id'
 gem 'httparty'
 gem 'rails_12factor', group: :production
 gem 'will_paginate'
-gem 'coveralls', '0.8.11', require: false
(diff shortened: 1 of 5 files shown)


lortza (approved):
Gorgeous! It's nice to see all of those Gemfile.lock dependencies gone-zo too!

Issue #373, opened 2023-08-30
Convert tags to badges


lortza:
Currently, touching the little x by a tag or list is difficult on mobile. I'd like to make that easier. My first idea is to expand the element by giving it an outline and to add more space between elements.

https://stackoverflow.com/questions/22390272/how-to-create-a-label-with-close-icon-in-bootstrap

PR #374, opened 2023-10-22
Add margin right to list and tag delete icons

lortza:
## Related Issues & PRs
Related to #373

## Problems Solved
When removing a movie from a tag or list on mobile, it is really hard to touch just that small circle-x icon without touching the neighboring item. This PR adds a little bit of right margin to make it a tiny bit easier.

I was originally thinking that we could turn the text + icon into bootstrap badges (#373) to give us a bigger touchable space, but I had forgotten that the list name itself is also a link that goes to that list page. I'm not really sure what I'd like this feature to do, so I went with just adding a little margin right for now to make this problem a tiny bit better.

While mucking around in the html, I cleaned up some rogue closing tags and fixed some blankspacing issues.

## Screenshots
| Before | After |
|-----|------|
| no margin | 5px margin right |
| Screen Shot 2023-10-22 at 11 19 57 AM |Screen Shot 2023-10-22 at 11 15 11 AM|

## Things Learned
* We have a lot of conditional logic in the view that could benefit from a refactor. 
* We could consider policies for general user access as well as view logic
* We are on a very old version of bootstrap, fyi

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -2,8 +2,10 @@
 <html>
   <head>
     <title><%= raw page_title %></title>
-    <link href='https://fonts.googleapis.com/css?family=Nixie+One|Open+Sans:400,600,700' rel='stylesheet' type='text/css'>
+    <!-- Bootstrap 3.0.2 -->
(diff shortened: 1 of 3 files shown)


lortza:
The build is failing for an unrelated chromedrivers issue. @mikevallano and i decided to merge this in and deal with the chromedrivers issue separately.

PR #375, opened 2023-10-22
Get tests passing again

lortza:
## Related Issues & PRs
Related to #374

## Problems Solved
While making a small cosmetic change in PR #374, the build broke on github. I ran tests on `master` locally several JS ones were failing. This means `master` is failing and the other PR was not the perp. Since the build was failing because of chromedrivers and we don't have that gem locally, I ran `bundle update` to so a sweeping cleanup of old gems hoping that yadda yadda yadda dependencies this would be resolved. After that update, the tests pass locally. I was hoping this would also do some magic to get the build to pass, but it did not.


mikevallano:
I mucked around with things and got the build passing.

I had to upgrade the ruby version 😞 , but since I did that, I ran `bundle update` on that branch, and here's the PR: https://github.com/mikevallano/tmdb-moviequeue/pull/376.

That means you'll have to install ruby `2.7.8` . But other than that I think that my PR is a good plan 🤷

lortza:
Sounds good to me. Let's close this one and get your PR merged.

mikevallano:
> Sounds good to me. Let's close this one and get your PR merged.

Done!

PR #377, opened 2023-11-24
Add my highest rated movies page

lortza:
## Problems Solved
* @lortza really wants to know what movies she has rated most-highly 
* we had no way of doing that easily
* This PR adds a simple page that displays all movies the current user has rated 8-10 all in one handy page
* includes pagination set to 20 items
* works on desktop and mobile
* cleans up the movie rating edit form by removing extra buttons and replacing them with a single `back` button and a link to the movie page 

## Screenshots
Screen Shot 2023-11-24 at 3 17 58 PM

| Navbar |mobile view|
|-----|-----|
| Screen Shot 2023-11-24 at 3 13 42 PM | Screen Shot 2023-11-24 at 3 36 07 PM |

## Things Learned
* `.where(value: 8..Float::INFINITY)` syntax requires an end argument in our version of rails
* you can do a fancy back button `link_to('Back', :back)` which will take you back to whatever page you came from instead of specifically saying which page to go back to

### DIFF
--- a/app/views/movies/_movie_rating.html.erb
+++ b/app/views/movies/_movie_rating.html.erb
@@ -5,7 +5,7 @@
         <icon class="fa fa-heart"></icon>
         <span class="pseudo-header"> Your Enjoyment:</span> <%= movie.ratings.by_user(current_user).first.value %>/10
         <%= link_to '<i class="fa fa-pencil"></i> Edit'.html_safe,
-                    movie_rating_path(movie, movie.ratings.by_user(current_user).first),
+                    edit_movie_rating_path(movie, movie.ratings.by_user(current_user).first),
                     id: "show_rating_link_movies_partial"
         %>
       </p>
--- a/app/views/ratings/edit.html.erb
+++ b/app/views/ratings/edit.html.erb
@@ -1,5 +1,6 @@
+<%= link_to('<- Back', :back) %>
+
 <h1>Editing Rating</h1>
+<p>for <%= link_to @movie.title, @movie %></p>
 
 <%= render 'form' %>
-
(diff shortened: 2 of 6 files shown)


config/routes.rb
```diff
@@ -7,6 +7,7 @@
 
   resources :users, only: [:show], as: :user do
     resources :lists
+    get 'ratings', to: 'users/ratings#index'
```
lortza, line 10: This gives us:
```
helper: user_ratings_path
url: GET  /users/:user_id/ratings(.:format) 	
controller: user/ratings#index
```

app/views/ratings/edit.html.erb
```diff
@@ -1,5 +1,6 @@
+<%= link_to('<- Back', :back) %>
```
lortza, line 1: This single back button takes the user back to where they were without making assumptions about where that was.

app/views/movies/_movie_rating.html.erb
```diff
@@ -5,7 +5,7 @@
         <icon class="fa fa-heart"></icon>
         <span class="pseudo-header"> Your Enjoyment:</span> <%= movie.ratings.by_user(current_user).first.value %>/10
         <%= link_to '<i class="fa fa-pencil"></i> Edit'.html_safe,
-                    movie_rating_path(movie, movie.ratings.by_user(current_user).first),
+                    edit_movie_rating_path(movie, movie.ratings.by_user(current_user).first),
```
lortza, line 8: I switched this to the `edit` path so the user doesn't have to stop by an annoying show page for a rating before getting to the edit functionality. This makes sense to me because the link to get to this rating page said "Edit".

mikevallano (approved):
🎉 Nice featch!

PR #382, opened 2023-11-28
381: upgrade rails to 6.0

### LINKED ISSUE #381: upgrade rails to 6.0

mikevallano:
## Related Issues & PRs
Closes #381

## Problems Solved
* Upgrades to Rails 6.0 as a step to get it up to date.

## Screenshots
Branch in production:
https://github.com/mikevallano/tmdb-moviequeue/assets/7945260/58b31922-a90f-46c3-837e-a1b6a22ba698

## Things Learned
* Circle was being a jerkle. 
* You can set ENV vars in Circle, which is one of the reasons things were failing.
* Rails didn't like auto loading from the root `lib` directory, so I moved stuff to `app/lib` instead.
* If you want to have constants start with an acronym (like `TV` or `API`), you have to add them in inflectors so that Rails doesn't complain about the constant name.


config/application.rb
```diff
@@ -1,50 +1,46 @@
-require File.expand_path('../boot', __FILE__)
+require_relative "boot"
 
 require "rails"
 # Pick the frameworks you want:
-require "active_storage/engine"
 require "active_model/railtie"
 require "active_job/railtie"
 require "active_record/railtie"
+require "active_storage/engine"
```
lortza, line 8: Do we want to require these (i'm assuming, large) chunks of rails code if we're not using them? I imagine they don't have cross-dependencies.

config/application.rb
```diff
@@ -1,50 +1,46 @@
...
+    # in config/environments, which are processed later.
     config.generators do |g|
-        g.test_framework :rspec,
-        fixtures: true,
-        view_specs: false,
-        helper_specs: false,
-        routing_specs: false,
-        controller_specs: true,
-        request_specs: false
-        g.fixture_replacement :factory_girl, dir: "spec/factories"
```
lortza, line 43: wow that was old. i imagine it was not presently working.

config/application.rb
```diff
@@ -1,50 +1,46 @@
...
+      helper_specs: false,
+      routing_specs: false,
+      controller_specs: true,
+      request_specs: false
+      g.fixture_replacement :factory_bot, dir: "spec/factories"
     end
-
-    # Do not swallow errors in after_commit/after_rollback callbacks.
-    config.autoload_paths += %W(#{config.root}/lib)
-    config.time_zone = 'Central Time (US & Canada)'
```
lortza, line 48: Do we want to set our time zone? Do you know what happens to our user experience if we don't?

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
+# Be sure to restart your server when you modify this file.
```
lortza, line 1: I believe the goal of this file is to uncomment each setting one at a time to make sure your app still works before flipping the main switch in the application.rb over to 6.0.

Once everything is working, you can delete this file.

lortza (commented):
Just a couple of questions to address.

config/application.rb
mikevallano: Ah, yeah I forget the impacts but I added it back. I didn't mean to remove it 😅

config/application.rb
mikevallano: > Do we want to require these

Good question.

Probably not, but I also don't feel like investigating them all to see what they do 😅 .

I commented out `active_storage` and `action_text` because I know we're not using them.

I _uncommented_ `action_cable` because I want to start adding that in, and I know I'll forget to check here to see if it's required.

PR #384, opened 2023-12-05
383: Upgrade Rails to 6.1, and Ruby to 3.0

### LINKED ISSUE #383: upgrade rails to 6.1
- Upgrade rspec. see https://github.com/rspec/rspec-rails#rspec-rails--
- Upgrade to ruby 3.0: https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html

mikevallano:
## Related Issues & PRs
Closes #383 

## Problems Solved
* Da lote won't have to jank around with ruby versions locally anymore

## Screenshots
Deployed the branch to production and it's working:

https://github.com/mikevallano/tmdb-moviequeue/assets/7945260/087452cd-7ab6-4b02-aeb0-c916dff39fbd

## Things Learned
* Relatively easy upgrade


.circleci/config.yml
```diff
@@ -3,12 +3,12 @@
...
 
 orbs:
-  browser-tools: circleci/browser-tools@1.4.0
+  browser-tools: circleci/browser-tools@1.4.6
 
 jobs:
   build:
     docker:
-      - image: cimg/ruby:2.7-browsers
+      - image: cimg/ruby:3.0.0-browsers
```
lortza, line 11: awwww yeah!

db/migrate/20231205190415_add_service_name_to_active_storage_blobs.active_storage.rb
```diff
@@ -0,0 +1,22 @@
+# This migration comes from active_storage (originally 20190112182829)
+class AddServiceNameToActiveStorageBlobs < ActiveRecord::Migration[6.0]
```
lortza, line 2: Just confirming -- you ran both of these migrations and they did nothing because we don't have the `active_storage_blobs` table. Is that correct?

lortza (approved):
Some minor questions. Nothing to block a merge.

Gemfile
```diff
@@ -1,13 +1,13 @@
...
+ruby '3.0.0'
 gem 'puma', '4.3.12'
-gem 'rails', '~> 6.0'
+gem 'rails', '~> 6.1.0'
 gem 'pg', '~> 1.1'
 gem 'sass-rails'
 gem 'uglifier', '>= 1.3.0'
 
-gem 'bigdecimal', '1.3.5'
+gem 'bigdecimal', '2.0.0'
```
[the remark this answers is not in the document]
mikevallano: good question. I looked into it and i don't think we need it. i removed it on the 7.1.2 branch and the tests passed, and i deployed to production and it seems like it's fine.

according to chatgpt:

> The bigdecimal gem is part of the Ruby standard library and provides a BigDecimal class for arbitrary-precision decimal arithmetic. This class is useful when working with numbers that require high precision, particularly in financial applications or any scenario where accurate decimal calculations are crucial.

> If you're not using or planning to use the BigDecimal class in your application, you may not need to explicitly include the bigdecimal gem in your Gemfile. In most cases, you can safely remove it, and your application will continue to work fine

So I'll keep it in this PR since there are other branches branched off of this one, and i'll remove it in the 7.1.2 branch

db/migrate/20231205190415_add_service_name_to_active_storage_blobs.active_storage.rb
mikevallano: omg, i just realized i added several comments as a review, but then failed to actually click the button to add the comments 🤦 .

yes, that's correct.

PR #419, opened 2024-04-17
Add dropdown to main nav using only css

lortza:
## Problems Solved
* give us basic css that we can use to make dropdowns in our main navbar without using javascript
* adds "lighter" and "darker" versions of our "secondary" color class
* consolidates (2) media queries to use the same screen width
* surfaces "mobile hidden" css by pulling it into its own class and moving it to the `shared` css
* lets us experiment with the idea of "bookmarked" content on our site for easy access (in this case, i've hardcoded our current TV shows to see if we like it)
* does not componentize this work to use in other places

## Technical considerations
I did end up commenting out the environment banner. It was interfering with the navbar positioning and as per our discussion, the juice wasn't worth the squeeze for our usage.

## Screenshots
Kapture 2024-04-17 at 10 41 29

## Things Learned
* css is always an interesting journey
* it's handy to use a checkbox input for toggling
* as long as we name our inputs differently, the toggling will work with multiple dropdowns on a page.
* hsl allows us to work around not having access to `lighten($color-secondary, 10%)` in plain css. ex: `hsl(48, 89%, 35%)`
* i keep forgetting that mobile does not have a hover state and end up trying to debug my css. 🤦‍♀️  _hopefully i learned it this time_.
* hardcoding is a great way to pilot an idea 😄

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,6 +1,6 @@
 <nav class="main-nav">
   <ul>
-    <li class="mobile-hidden"><%= link_to "Home", root_path %></li>
+    <li class="mobile-hidden"><%= link_to 'Home', root_path %></li>
     <% if current_user.present? %>
       <li><%= link_to 'My Movies', movies_path %></li>
       <li><%= link_to 'My Lists', user_lists_path(current_user) %></li>
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,10 +1,10 @@
 <nav class="main-nav">
   <ul>
-    <li><%= link_to 'Home', root_path %></li>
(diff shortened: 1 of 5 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -1,19 +1,27 @@
 <nav class="main-nav">
   <ul>
-    <li><%= link_to 'Home', root_path %></li>
+    <li class="mobile-hidden"><%= link_to "Home", root_path %></li>
```
lortza, line 3:
```suggestion
    <li class="mobile-hidden"><%= link_to 'Home', root_path %></li>
```

lortza:
Moving back to draft to figure out position problem.Try:
* Compare to responsive nav bar example 
* remove padding from nav container and make fixed width
* see why mobile is so much smaller. Look at actual measurements

mikevallano (commented):
noice!

PR #437, opened 2024-09-17
Link to actor bio or episode page from TV pages

lortza:
## Problems Solved
When looking at actor credits from TV shows, we didn't like needing to go through the actor's movies list in order to get to the bio. We fixed this for the movie actors, but not the TV actors. This PR fixes it for the TV actors.

1. Links to main cast member's bio page instead of movies search page on the:
    * Series `:show` page
    * Season `:show` page
    * Episode `:show` page

2. Links to guest star's episodes listings for this show on the Episode `:show` page
    * I would like to have linked consistently to the episodes listings page from the Episode `:show` page, but weirdly, there isn't any data for the Cast actors. 
    * I chose to _not_ link to the Guest actors' bio pages because I usually want to know how many episodes of this series the person has been in (kind of like Jeffrey Combs).
4. Separates Cast actors from Guest actors because the links are different on the Episode `:show` page

### DIFF
--- a/app/views/shared/_tv_series_profile_content.html.erb
+++ b/app/views/shared/_tv_series_profile_content.html.erb
@@ -38,7 +38,7 @@
           <div class="headshot-container">
             <%= headshot_for(actor) %>
             <div class="name-block">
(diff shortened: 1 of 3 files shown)


lortza:
This is handled in #460

PR #452, opened 2024-12-15
Edit tv series viewings

lortza:
## Related Issues & PRs
Closes #488

## Problems Solved
* allows users to delete viewing records
* allows users to edit viewing records
* orders navbar dropdown items by title alpha ascending

## Screenshots
The button to mark a current show as finished is gone and now we're offering an edit link instead
Screenshot 2024-12-15 at 1 43 16 PM

This is the edit form
Screenshot 2024-12-15 at 1 43 49 PM

This is the navbar with shows ordered by title
Screenshot 2024-12-15 at 2 06 15 PM

## Approach
There were several ways to do this work -- all of them seem to have tradeoffs. I settled on using `turbo: false` for the edit form and having the controller handle the different behaviors based on format. The behaviors are:
* when adding an item to the watching list via the toggle button, turbo handles DOM updates and the user stays on the tv series page
* when removing an item from the watching list via the toggle button, turbo handles DOM updates and the user stays on the tv series page
* when editing an item via the edit form, the html does a redirect back to the index page

### DIFF
--- a/app/views/tv_series_viewings/_navbar_dropdown.html.erb
+++ b/app/views/tv_series_viewings/_navbar_dropdown.html.erb
@@ -5,7 +5,7 @@
     <ul class="dropdown-items">
       <li><%= link_to 'See Viewing History', tv_series_viewings_path %></li>
       <hr>
-        <% current_user.tv_series_viewings.active.each do |show| %>
+        <% current_user.tv_series_viewings.active.order(title: :asc).each do |show| %>
           <li><%= link_to show.title, show.url %></li>
         <% end %>
     </ul>
--- a/spec/factories/tv_series_viewings.rb
+++ b/spec/factories/tv_series_viewings.rb
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :tv_series_viewing do
     user
-    title { FFaker::HipsterIpsum.phrase }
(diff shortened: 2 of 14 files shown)


app/views/tv_series_viewings/_navbar_dropdown.html.erb
```diff
@@ -5,7 +5,7 @@
     <ul class="dropdown-items">
       <li><%= link_to 'See Viewing History', tv_series_viewings_path %></li>
       <hr>
-        <% current_user.tv_series_viewings.active.each do |show| %>
+        <% current_user.tv_series_viewings.active.order(title: :asc).each do |show| %>
```
lortza, line 8: Tossed in a quick order here because we wanted it

spec/factories/tv_series_viewings.rb
```diff
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :tv_series_viewing do
     user
-    title { FFaker::HipsterIpsum.phrase }
+    sequence(:title) { |n| "TV Series #{n}" }
```
lortza, line 4: The ipsum generated here has some icky symbols. I went with a simple sequence here instead.

PR #455, opened 2025-01-02
Add streaming service providers

lortza:
## Problems Solved
* Since Hooplah is not casting properly, I found a new provider (Kanopy) and included it in our list of providers
* We have Peacock Premium (and maybe Premium Plus) for a brief moment, so I'm including it in the list.
* VUDU has changed. It has a free-with-ads pay model now, so I'm commenting it out
* VUDU free has rolled into Fandango, so that data is updated
* I'm using api ids to match services instead of name matching now
* I've incorporated a `display_name` so we're not beholden to the API `provider_name`

## Things Learned
* I just started using the VS code plugin called "Code Spell Checker" and that caught a typo for us
* It would be neat if TMDB had an API endpoint for a service providers index, but it does not
* It would also be neat if it were easier to add/remove streaming providers, but at least right now we can comment them out of the data hash

### DIFF
--- a/spec/services/streaming_service_provider_data_service_spec.rb
+++ b/spec/services/streaming_service_provider_data_service_spec.rb
@@ -20,43 +20,36 @@ RSpec.describe StreamingServiceProviderDataService do
       it 'returns our default providers with a pay_model of "try"' do
         allow(Tmdb::Client).to receive(:request).and_return(nil)
 
-        results = StreamingServiceProviderDataService.get_providers(**movie_args)
-        aggregate_failures "all default providers appear with 'try'" do
-          expect(results.find {|r| r[:name] == 'Netflix'}[:pay_model] ).to eq('try')
-          expect(results.find {|r| r[:name] == 'Amazon Prime Video'}[:pay_model] ).to eq('try')
-          expect(results.find {|r| r[:name] == 'Amazon Video'}[:pay_model] ).to eq('try')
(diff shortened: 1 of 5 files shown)


spec/services/streaming_service_provider_data_service_spec.rb
```diff
@@ -20,43 +20,36 @@
...
       end
     end
 
     describe 'when the API does return results' do
+      let(:netflix_api_data) { {logo_path: "/5OAb2w.jpg", provider_id: 8, provider_name: "Netflix", display_priority: 26} }
+      let(:amazon_api_data) { {logo_path: "/5OAb2w.jpg", provider_id: 10, provider_name: "Amazon Video", display_priority: 26} }
+      let(:amazon_prime_api_data) { {logo_path: "/5OAb2w.jpg", provider_id: 9, provider_name: "Amazon Prime Video", display_priority: 25} }
+      let(:youtube_api_data) { {logo_path: "/xL9SUR.jpg", provider_id: 192, provider_name: "YouTube", display_priority: 51} }
+      let(:fandango_api_data) { {logo_path: "/peURlL.jpg", provider_id: 7, provider_name: "Fandango At Home", display_priority: 4} }
+
```
lortza, line 35: These new `let`s clean up a lot of redundant data hashes throughout the spec.

mikevallano: noice!

mikevallano (approved):
awww yeah! thanks for doing this.

PR #456, opened 2025-01-04
Simplify streaming provider views

lortza:
## Problems Solved
* Reduces clutter for streaming services
* Only shows free options if free options are available
* Does not show "try" options on mini modal
* Gets us 1 step closer to user-configurable streaming service providers

## Screenshots
| Before | After |
|-----|-------|
| Screenshot 2025-01-04 at 4 07 54 PM | Screenshot 2025-01-04 at 4 06 29 PM |
| Screenshot 2025-01-04 at 4 08 04 PM | Screenshot 2025-01-04 at 4 06 40 PM |
| Screenshot 2025-01-04 at 4 09 16 PM | Screenshot 2025-01-04 at 4 09 35 PM |

## Things Learned
* I chose to convert our streaming service providers data into an active hash to keep getting practice using this data type
* This will allow us to easily create a `users_streaming_service_providers` table with the `id`s from each active hash object so users can configure their service providers via a gui in a subsequent PR
* I used fancy css to handle the comma separation

## Checks
- [x] update all relevant specs
- [x] run whole test suite on `master` to get number of feature test failures: `130 examples, 117 failures, 13 pending`
- [x] run whole test suite on this branch to get number of feature test failures: `130 examples, 117 failures, 13 pending`


app/assets/stylesheets/media-show-page.css
```diff
@@ -30,22 +30,11 @@
...
 .manage-screenings .material-symbols-outlined.edit-note {
   position: relative;
   top: 7px;
 }
 
-.txt-sm.availability {
-  font-size: 10px;
-  margin-top: 20px;
-}
-
```
lortza, line 48: Replaced this with a simple utility class that uses tailwind specs: https://tailwindcss.com/docs/font-size

app/helpers/movies_helper.rb
```diff
@@ -58,15 +58,6 @@ def movie_genres_display(movie)
...
 
-  def display_pay_model_icon(movie_pay_model)
-    case movie_pay_model
-      when 'try' then 'search_check'
-      when 'free' then 'emoji_emotions'
-      when 'rent' then 'paid'
-      when 'buy' then 'shopping_bag'
-    end
-  end
-
```
lortza, line 69: We don't need this anymore since we're only rendering these icons in 1 partial.

app/views/movies/_modal_content.html.erb
```diff
@@ -9,8 +9,7 @@
     </div>
     <div class="watch-info">
       <%= render 'movies/screenings', movie: movie %>
-      <%= render "movies/mini_modal/streaming_service_providers",
-          streaming_service_providers: movie.streaming_service_providers %>
```
lortza, line 13: I've condensed display logic across the app so we don't need 2 different (but very similar) partials anymore

PR #458, opened 2025-01-05
Let users configure streaming service providers

lortza:
## Problems Solved
* Not all users have the same streaming services, so this PR offers user selection of services instead
* Editing the available providers required a PR & deploy, so this PR offers a GUI experience
* Doubles the amount of available streaming service provider data

## Things Learned
*


app/entities/streaming_service_provider.rb
```diff
@@ -1,108 +1,245 @@
...
-    where(pay_model: "rent")
-  end
+  field :search_url, default: ''
 
-  def self.buy
-    where(pay_model: "buy")
-  end
-
-  def self.maybe_available
-    where(pay_model: "not_found")
```
lortza, line 17: These methods were dubious to begin with. Now we're 100% not using them.

Issue #459, opened 2025-02-23
Separate actor search from movies-for-actor search


lortza:
This is related to #380.

The current controller action in the `app/controllers/tmdb_controller.rb` is called `actor_search`. 
```ruby
  def actor_search
    if params[:actor].present?
      @results = MovieDataService.get_movies_for_actor(
        actor_name: params[:actor],
        page: params[:page],
        sort_by: params[:sort_by]
      )
    end
  end
```

We can rename that to `movies_for_actor_search` and then use the endpoint name `actor_search` for a method that locates the actor. This can then link directly to the bio
```ruby
def actor_search
    if params[:actor].present?
      person_data = Tmdb::Client.request(:person_search, query: params[:actor])[:results]&.first
      if person_data.present?
        redirect_to actor_more_path(actor_id: person_data[:id])
      else
        flash.now[:alert] = "No results found. Please try again."
        render :actor_search, status: :unprocessable_entity
      end
    end
  end
```

PR #460, opened 2025-02-23
Clean up actor tv credits

### LINKED ISSUE #380: Search actor by id
This little 80s gem https://youtu.be/8cMpCr-SN7U?feature=shared is in parks & rec Brandanowitz's credits http://www.flicksonlists.com/tmdb/actor_search?actor=Paul+Schneider

And that's hilarious, but I doubt he was a director for Disney in 1988. Also, his bio is some other person. http://www.flicksonlists.com/tmdb/actor_more?actor_id=543495

I think the trouble is that we do name searches in some places instead of by id.  My guess is we had the id on the parks & rec cast page, then used his name to search for movies -- got the wrong guy -- then followed that wrong id to the wrong bio. 

If I recall from all of the refactor work, one does not simply fix this. We reuse the searches from different locations, for example, allowing a user to type in an actor name or arriving on that page via another credit.

lortza:
## Related Issues & PRs
Related to #380

## Problems Solved
* Streamlines the tv "appearance details" link on an actor bio by using the character name as a link to the character appearances
* Replaces the link to actor credits from `actor_search_path` to `actor_more_path`, which is the bio page. Which partially closes #380.
* I was very tempted to also make our actor search result redirect to the bio page, but I broke it out of #380 and moved it to its own issue #459.

## Screenshots
My goal here for these visual changes is to reduce visual clutter and annoying line wraps when trying to visually scan a list of TV credits on my phone. This change makes each line shorter and reduces the need for a line wrap.
 
|Before | After |
|----|-----|
|Screenshot 2025-02-23 at 9 21 15 AM | Screenshot 2025-02-23 at 9 21 01 AM | 

Sometimes an actor is on a show but does not have a character name. For example, Jason Sudeikis was on SNL and doesn't have an individual credit for each skit character he plays. In those cases, we just render the word "Episodes" -- much like the original behavior with the phrase "Appearance Details".

### Due Diligence Checks
- [ ] wrote new specs for this work
- [x] ran non-feature specs locally
- [x] tested it in the browser locally
- [x] deployed branch to prod and browser tested it (before merging to `master`)

### DIFF
--- a/app/views/shared/_tv_series_profile_content.html.erb
+++ b/app/views/shared/_tv_series_profile_content.html.erb
@@ -39,7 +39,7 @@
           <div class="headshot-container">
             <%= headshot_for(actor) %>
             <div class="name-block">
-              <p class="actor"><%= link_to "#{truncate(actor.name, length: 18, escape: false)}", actor_search_path(actor: I18n.transliterate("#{actor.name}")) %></p>
+              <p class="actor"><%= link_to "#{truncate(actor.name, length: 18, escape: false)}", actor_more_path(actor_id: actor.actor_id) %></p>
               <p>as <%= truncate(actor.character_name, length: 18, escape: false) %></p>
             </div>
           </div>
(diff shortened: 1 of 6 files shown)


lortza:
I fixed that syntax error and we should be good to merge now.

PR #461, opened 2025-02-26
Add main cast content to episode credits page

lortza:
## Problems Solved
**TL;DR:** it's imperfect, but an improvement for most cases.

It is very weird to go to the TV episode credit page for a main cast actor and see nothing.
BEFORE:
Screenshot 2025-02-26 at 10 44 39 AM

This is because of the way TMDB handles its episode data for these characters. If we were storing all of this data in our own db, it would be easy to query, but this would be a lot of API hits to get everything for this page. So I'm making this small improvement where we acknowledge that the person was in the main cast for the seasons that they were.
AFTER:
Screenshot 2025-02-26 at 10 44 12 AM

The Natalie (Monk) situation is a little weird. Even though this character was introduced halfway through season 3, she's not credited as main cast until season 4. The end result is that we were only seeing a handful of episodes on this credits page when we know she's in every episode from season 4 through 8. I was able to work around this weirdness by comparing the actor's id with the list of ids in the series cast with a subsequent API call. 
Screenshot 2025-03-03 at 10 43 29 AM

But it gets harder when we have a Sharona (Monk) or Kes (Voyager) situation. Both Sharona and Kes are replaced as main cast in roughly season 3 of their respective tv shows, so they don't appear in the TMDB series main cast data. This means, we're only seeing the credits for when they were non-cast on the show. For example, Sharona was in the main cast on Monk for seasons 1, 2, and half of 3. Then she appeared once in season 8. The credits only show those overlap seasons (3 and 8). Similarly, Kes was main cast for seasons 1-3 and partial season 4. But we're only seeing the overlap episodes in season 4 and the single episode in season 6 when she came back as a guest.

Screenshot 2025-02-26 at 10 47 37 AM
Screenshot 2025-03-03 at 10 51 43 AM

I may be able to resolve that with a lot of fancier API calls, but I'm not sure we care enough to do that. For that reason, I'm choosing not to solve for that particular situation in this PR.

## Data Patterns
First of all, the data is imperfect, so I made the best assumptions I could with what I was seeing. In order to make this `is_main_cast?` logic, I looked at data for different character types across a few shows. Here is a sample of the data for the Monk characters:
```
Sharona has 
  episodes: select from seasons 3 and 8
  seasons: Specials, 1, 2, 8
Natalie has 
  episodes: select from season 3
  seasons: 4, 5, 6, 7, 8
Randy has
  episodes: []
  seasons: Specials, 1, 2, 3, 4, 5, 6, 7, 8
Stottlemeyer has
  episodes: []
  seasons: Specials, 1, 2, 3, 4, 5, 6, 7, 8
Guest star 1 has
  episodes: one from season 3
  seasons: 3
Guest star 2 has
  episodes: one from season 3
  seasons: []
```

## Merge Checklist
- [x] I have written specs for this work
- [x] I ran specs locally with `bundle exec rspec spec --tag ~@feature`
- [x] I have browser tested this locally for different scenarios
- [ ] I have deployed the branch to production for testing

### DIFF
--- a/app/entities/tv_actor_credit.rb
+++ b/app/entities/tv_actor_credit.rb
@@ -18,7 +18,7 @@ class TVActorCredit
   def self.parse_record(record)
     seasons = TVCreditSeason.parse_records(record[:media][:seasons]).presence if record[:media][:seasons].present?
     episodes = TVEpisode.parse_records(record[:media][:episodes]).presence if record[:media][:episodes].present?
-    episodes_by_number = episodes.sort_by { |e|e.episode_number } if episodes.present?
+    episodes_by_number = episodes.present? ? episodes.sort_by { |e|e.episode_number } : []
     new(
       actor_name: record[:person][:name],
       actor_id: record[:person][:id],
@@ -30,5 +30,9 @@ class TVActorCredit
       show_name: record[:media][:name],
       seasons: seasons
     )
+  end
+
+  def in_main_cast?
+    @character.present? && @episodes.empty?
   end
 end
--- a/app/entities/tv_actor_credit.rb
+++ b/app/entities/tv_actor_credit.rb
@@ -16,7 +16,7 @@ class TVActorCredit
   end
 
   def self.parse_record(record)
-    seasons = TVCreditSeason.parse_records(record[:media][:seasons]).presence if record[:media][:seasons].present?
+    seasons = TVCreditSeason.parse_records(record[:media][:seasons]).sort_by {|s| s.name}
     episodes = TVEpisode.parse_records(record[:media][:episodes]).presence if record[:media][:episodes].present?
     episodes_by_number = episodes.present? ? episodes.sort_by { |e|e.episode_number } : []
     new(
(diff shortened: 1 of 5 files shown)


app/entities/tv_actor_credit.rb
```diff
@@ -16,9 +16,9 @@ def initialize(show_name:, show_id:, actor_name:, actor_id:, character:, seasons
   end
 
   def self.parse_record(record)
-    seasons = TVCreditSeason.parse_records(record[:media][:seasons]).presence if record[:media][:seasons].present?
+    seasons = TVCreditSeason.parse_records(record[:media][:seasons]).sort_by {|s| s.name}
     episodes = TVEpisode.parse_records(record[:media][:episodes]).presence if record[:media][:episodes].present?
-    episodes_by_number = episodes.sort_by { |e|e.episode_number } if episodes.present?
+    episodes_by_number = episodes.present? ? episodes.sort_by { |e|e.episode_number } : []
```
lortza, line 21: In both of these cases, I want an empty array instead of nil.

app/views/tmdb/actor_credit.html.erb
```diff
@@ -3,7 +3,17 @@
...
+<br>
+
+<% if @credit.in_main_cast? %>
+  <p class="mt-10"><%= @credit.actor_name %> was in the main cast as <%= @credit.character %> for <%= @credit.seasons.size %> seasons.</p>
+  <ul>
+    <% @credit.seasons.sort_by {|season| season.season_number}.each do |season| %>
+      <li><%= link_to season.name, tv_season_path(show_id: @credit.show_id, season_number: season.season_number) %></li>
+    <% end %>
+  </ul>
+<% end %>
```
lortza, line 16: Screenshot 2025-02-26 at 10 44 12 AM

PR #463, opened 2025-03-01
462: Clean up create logic in the controller. Add unique index

### LINKED ISSUE #462: Prevent dupe listings
We've seen a couple instances of dupe listings lately.

Looking at the data, it looks like a race condition:

```ruby
#<Listing:0x00007f9c6348d490
  id: 12533,
  list_id: 5,
  movie_id: 1127,
  created_at: Sun, 23 Feb 2025 23:21:06.950117000 UTC +00:00,
  updated_at: Sun, 23 Feb 2025 23:21:06.950117000 UTC +00:00,
  user_id: 5,
  priority: 3>,
 #<Listing:0x00007f9c6348d710
  id: 12534,
  list_id: 5,
  movie_id: 1127,
  created_at: Sun, 23 Feb 2025 23:21:06.954739000 UTC +00:00,
  updated_at: Sun, 23 Feb 2025 23:21:06.954739000 UTC +00:00,
  user_id: 5,
  priority: 3>]
```

- clean up the form and controller
- add a unique index

mikevallano:
## Related Issues & PRs
Closes #462 

## Problems Solved
* prevents dupe listings

## Screenshots
https://github.com/user-attachments/assets/7caf2ca1-9acf-4b9c-8d70-56f1368a49cb

## Things Learned
* Not sure _why_ this is happening because the UI simply selects a list to add the movie to.
But apparently sometimes an additional request is made, and even though we have a unique validation, it's a race condition. So adding the unique index will prevent it from happening, and then the controller will rescue and the user will be none the wiser.


db/migrate/20250301135740_remove_redundant_listing_indexes.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveRedundantListingIndexes < ActiveRecord::Migration[7.1]
+  def up
```
mikevallano, line 2: since we're adding a composite index on [:movie_id, :list_id], the db can use that index and we don't need the individual indexes

PR #501, opened 2026-05-04
Add streaming providers

lortza:
## Problems Solved
We were not able to see Paramount+ info on movie and tv listings.  As I suspected, we just didn't have the most up-to-date provider info stored in our data. 

## This PR
* Adds the Paramount Plus subscription data so these options will appear for users to select in their settings
* Adds a couple of other providers as well

## Screenshots
Before: only showing Try options
Screenshot 2026-05-04 at 9 40 09 AM

After: showing Paramount+ options
Screenshot 2026-05-04 at 9 40 21 AM

## Things Learned
When I built this originally, I made the streaming service provider data static so we could get an MVP up and running without thinking about database table requirements. So, the process for this fix was for me to manually search for content that I know is on Paramount then `pry` to see which providers are returned by the API. I then added them to the array. We may want to consider changing the implementation strategy to a table that gets updated either upon specific request calls or with a cron job so that we a least have access to new providers as they form in the API -- instead of us just wondering why nothing is appearing.

### DIFF
--- a/app/entities/streaming_service_provider.rb
+++ b/app/entities/streaming_service_provider.rb
@@ -210,6 +210,14 @@ class StreamingServiceProvider < ActiveHash::Base
       search_url: "https://www.peacocktv.com/watch/search?q="
     },
     {
+      id: "a038107b-d7bd-4ff9-bcd2-6fdec3028ea4",
+      display_name: 'Philo',
+      tmdb_provider_name: "Philo",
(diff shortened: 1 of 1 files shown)


lortza:
The build is failing because of something unrelated and I don't have circle permissions to rerun.

mikevallano (approved):
🚀

PR #509, opened 2026-06-16
Upgrade From Rails 7.2.3.1 to 8.0.0

lortza:
## Summary
Upgraded Rails from 7.2.3.1 to 8.0.5 and migrated from Sprockets to Propshaft asset pipeline. The work for this upgrade PR cost $21.73 in Claude Sonnet 4.5 tokens with a lot of human intervention.

### Rails Upgrade
- Upgraded from Rails 7.2.3.1 to 8.0.5
- Updated `config.load_defaults` to 8.0
- Adopted all Rails 8.0 framework defaults:
  - `active_support.to_time_preserves_timezone = :zone`
  - `action_dispatch.strict_freshness = true` 
  - `Regexp.timeout = 1` (security improvement)

### Asset Pipeline Migration (Sprockets → Propshaft)
- Replaced `sprockets-rails` with `propshaft` gem
- Removed `require "sprockets/railtie"` from application.rb
- Updated layout to use `stylesheet_link_tag :app` (Propshaft convention)
- Cleaned up application.css (removed Sprockets manifest directives)
- Removed Sprockets-specific config from development/production environments
- Kept assets.rb initializer (harmless with Propshaft, ignored)

### Gem Cleanup
- **Removed deprecated/unnecessary gems:**
  - `uglifier` (not needed with Propshaft, no JS minification)
  - `hotwire-rails` (redundant, we already have turbo-rails + stimulus-rails)
  - `rails_12factor` (Rails 5+ has this built-in)
  - `webdrivers` (deprecated, selenium-webdriver handles this now)
  - `sass-rails` (migrated to Propshaft)

- **Pinned gem versions for stability:**
  - `devise '~> 5.0'` (was unpinned, now locked to 5.x)
  - `redis "~> 4.0"` (was unpinned, now locked to 4.x)

### Configuration Updates
- Restored TV acronym in inflections initializer (was lost during app:update)
- Updated all environment configs to Rails 8.0 defaults
- Preserved custom configs:
  - SendGrid SMTP settings in production
  - letter_opener in development
  - Custom generator settings (RSpec, FactoryBot)
- Enabled `force_ssl = true` in production (Rails 8 default)
- Added health check silence for `/up` endpoint

### Testing
- Fixed rails_helper.rb (removed webdrivers require)
- All non-feature tests passing: **464 examples, 0 failures, 2 pending**
- Feature tests remain disabled (as configured in CI)

## Key Decisions

[...]

PR #520, opened 2026-07-29
Convert CI from Circle to GitHub Actions

lortza:
## Problems Solved
This PR moves the CI checks from Circle to GitHub actions. Each of the checks is in a separate job that runs in parallel, which saves time. I didn't set up any requirements for these to pass before a PR can be merged. If we want to, it's easy to do in github settings.

In the process, I added the `tmdb_api_key` to the github secrets via https://github.com/mikevallano/tmdb-moviequeue/settings/secrets/actions

Note: 
This work was initially released with the Rails 8 upgrade #509, but since I reverted that PR, I am releasing this separately.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+  #       uses: actions/checkout@v6
+
+  #     - name: Set up Ruby
+  #       uses: ruby/setup-ruby@v1
+  #       with:
+  #         ruby-version: .ruby-version
+  #         bundler-cache: true
+
+  #     - name: Scan for common Rails security vulnerabilities using static analysis
+  #       run: bin/brakeman --no-pager
```
lortza, line 23: if we want to add brakeman to ci, we can by uncommenting it

.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+  #       with:
+  #         ruby-version: .ruby-version
+  #         bundler-cache: true
+  #         cache-version: 1 # Increment this to clear cache if needed
+  #     - name: Unfreeze Bundler for Dependabot
+  #       if: github.actor == 'dependabot[bot]'
+  #       run: bundle config set frozen false
+
+  #     - name: Lint code for consistent style
+  #       run: bundle exec standardrb
```
lortza, line 63: If we want to run a ruby linter in ci, we can by uncommenting this. We'll also need to add the `standard` gem and run it locally to pick up any linting changes.

